PR #1, opened 2022-02-09
Pin protocol proposal

Gozala:
This is current draft of the pin protocol proposal.

### DIFF
--- a/Pin.md
+++ b/Pin.md
@@ -1,0 +1,209 @@
+# ⚠️ Disclaimer ⚠️
+
+> Since writing this I have realized that while incermental writing approach described here would work well in centralized system (single writer) it will be problematic in decentralized system where multiple actors may be doing concurrent updates e.g. if  web3 application uses web3.storage service from multiple clients they would fail to transact or will have to coordinate updates (version & index).
+>
+> I think we could do better by using grow-only sets when patching the pin object and only coordinate pin update between `Pinned` states. That way `Transient` pins could be updated concurrently without coordination & only coordinate updates on `Pinned` pins. 
+
+# Pins
+
+Following is an IPLD Schema representation of the "pin" objects (or whatever we want to call it) which:
+
+1. Can be in "tranisent" or "pinned" state
+    - To allow incermental updates through series of transactions.
+    - To have `tranisent` representation series of between udates.
+3. Are identified by an [ed25519][] public key. Therefor they can represent
+    - IPNS names
+    - [did:key][] identifiers
+    - Actors in [UCAN][] authorization
+
+
+This design would address several problems in .storage services:
+
+### Large uploads
+
+Large uploads that span multiple CAR files would gain a first class representation. Client application will be able to self issue new "pin" identifier and through incremental transactions amend it's state by uploading DAG shards via CAR files. Each transaction would amend `Pin` object with additional DAG shard head(s) followed by a final trasaction changing `Pin` from `Transient` state to `Pinned` state pointing to
+a desired DAG root cid (that was provided in one of the transaction).
+
+This way would allow .storage service to list not only succesful, but also "in progress" uploads (pins). Additional metadata could also be used to provide domain specific information about the status. E.g. applications built on top of web3.storage could utilize this to provide human readable description along with domain specific status `code`.
+
+
+### IPNS
+
+.storage services could directly map pins to corresponding IPNS names, making it possible to access arbitrary uploads / pins through an IPNS resoultion.
+
+Pin status could be used to decide when to propagate pin updates through the network e.g. sevice could choose to only announce only pinned states.
+
+### did:key
+
+.storage service could also provide interface for accessing content under `did:key` that correspond to a given keys. Basically we can build IPNS like system except with delegated publishing through UCANs before integrating that into IPNS.
+
+### UCAN
+
+By representing pins as first class objects identified by `did:key` they become actors in UCANs delegated capabilties system.
+
+.storage user could issue delegated token for specific `Pin` object and excercise that capability to update given `Pin` object or delegate that capability to another actor in the system.
+
+
+## Schema
+
+Following is an IPLD schema definition for the `Pin` object. 
+
+> 💭 One one hand we would like to specify enough structure to be able to make sense of it in applications (be it .storage or it's clients), but on the other hand boxing actual DAG just to give it a "status" info seems awkward.
+
+
+```ipldsch
+-- Pin represents (IPNS) named pointer to a DAG that
+-- is either in "transient" state that is partially
+-- or fully "pinned" state. In both cases it is
+-- anotated pointer to DAG head(s).
+type Pin union {
+    Transient "transient"
+    Pinned "pinned"
+} representation inline {
+  discriminantKey "status"
+}
+
+-- Represents partially pinned DAG. Think of it as
+-- dirty tree in the git, head points to previous
+-- revision.
+-- Please note: Even though pin links to a previous
+-- revision of the pin there is does not imply it is
+-- pinned (you would need to include that link in the
+-- index for that)
+type Transient struct {
+  -- Link to a previous pin revision in "pinned" state
+  head &Pinned
+  -- Set of DAG roots that next pinned state will be
+  -- comprised of.
+  -- Please note that providing blocks under DAG
+  -- happens out of band meaning that DAG under the
+  -- link could be partial.
+  links [Link]
+}
+
+-- Represents fully pinned DAG with some metadata.
+-- Please note that fully pinned DAG does not imply
+-- that full DAG is pinned, but rather provided
+-- subdag
+type Pinned struct {
+    -- Root representing current state of the pin.
+    root Link
+    -- Previous version of this pin (not sure what
+    -- would genesis block pint to maybe we need
+    -- a special genesis variant of "Pin" union)
+    head &Pinned
+    -- We have links to all the relevant sub-DAGs
+    -- because `root` may not be traversable e.g
+    -- if it is encrypted. By providing links service
+    -- can traverse it and pin all the relevant blocks
+    -- even when it can't make sense of them.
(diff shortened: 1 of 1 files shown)


Gozala:
> how do we enforce `Client SHOULD NOT reuse keypair across multiple upload sessions.` ? block pub keys in the kv ?

To be clear it's not inherently wrong to reuse keys, it's just going to replace existing upload associated with key with another, which I don't think what we want right now. I do not think we need to enforce that we should just generate new keypair for every upload until we're further along in the mutable land

Revision.md
```diff
@@ -0,0 +1,238 @@
...
+will be referenced as CAR roots and point to the DAG-CBOR encoded `Transaction` object with couple of
+nuances:
+
+1. Transaction `links` may link to CAR CID which is to be intepreted as a set of `links` for all the blocks contained in the corresponding CAR.
+
+   > **Note:** We do not currently have CAR IPLD codec. Idea is to iterate on this and define spec based on lessons learned.
+
+2. If `links` field are omitted from transaction object that implies links to all the blocks of this CAR (except transaction blocks).
+
+> 💔 I do not like "implicit links" as that seems impractical in case of multiple transactions. Even for a single transaction case `Transaction` may want to link a DAG known to be available at the destination e.g. CID of the previous revision.
```
mikeal, line 221: this is pretty easy to distinguish, because CAR CID’s will have a specific codec, so all other CID’s would be links to previous revision.

Revision.md
```diff
@@ -0,0 +1,238 @@
+
+# Revision
```
mikeal, line 2: I think “session” may be a better word to use than “revision”

Revision.md
```diff
@@ -0,0 +1,238 @@
...
+This design would address several problems in .storage services:
+
+### Large uploads
+
+Large uploads that span multiple CAR files would gain a first class representation via `Revision` objects. Client application wishing to upload large file (or any other DAG) will be able to accomplish that by:
+
+
+1. Self issuing a new `Revision` identifier (and corresponding UCAN) by generating new [ed25519][] keypair.
+1. Submitting concurrent `Patch` transactions (in [CAR][] format). Each transaction will contain DAG shards, subset of blocks that would feet upload quota.
+1. Finalizing upload by submitting `Commit` transaction (in [CAR][] format), setting `root` of the `Revision` to a `CID` of the large file.
```
mikeal, line 24: s/“of the large file”/“in the final CAR”

Revision.md
```diff
@@ -0,0 +1,238 @@
...
+## Schema
+
+Following is an [IPLD schema][] definition for the `Revision` object. 
+
+
+```ipldsch
+-- Revision represents (IPNS) named pointer to a DAG that
+-- is either in "draft" or fully "release" state.
+-- In both cases it is anotated pointer to a DAG.
+type Pin union {
```
mikeal, line 56: i think you already know this, but let’s get rid of the work “Pin” anywhere in the document :)

for this structure specifically, i think you can call it “Session”

Gozala:
Looks like ‘revision’ name is also controversial, so I propose to bikeshed on this further here. While renaming Pin to Revision I have considered following alternatives:

1. `Ref` protocol (Personally not a fan but idea is to update refs so maybe it works)
2. `IPTM (Inter Planetary Transactional Memory)` - There is quite a bit of overlap with STMs so maybe we could lean into it more.
3. `Record` protocol - Too generic
4. `Account protocol` - I actually like this, but I also recognize that the name might be confusing especially when used in context of blockchains etc…
5. `Chronicle` - Emphasizing fact that it’s a chain of events.

@mikeal suggested `Session protocol`, which I find misleading, because even though for file uploads it will act like a session, I do hope that it would also enable interesting uses for mutability in web3.storage where it will not correspond to session, but rather to a revision history of the mutable DAG.

I also would like to emphasize transactionalitiy more than anything. So here are few other ideas

6. Update protocol
7. Mutation protocol
8. Transactional update protocol
9. Transaction protocol
10. IPTS (Inter Planetary Transaction System)

Please (ranked) vote and/or propose alternatives.

mikeal:
I like IPTM :)

the-simian:
IPTM and IPTS make the most sense to me, with IPTM being my favorite between the two.

Pin.md
```diff
@@ -0,0 +1,323 @@
...
+1. Transactional update protocol for these `Documents`.
+1. Document publishing protocol and associated concensus algorithm.
+
+
+### Document Model
+
+Document represents a view of the DAG in time, uniquely identified by [ed25519][] public key. It is state derived from set of operations that were authorized via corresponding private key (through [UCAN][]).
+
+> Because documents are identified by [ed25519][] public key, that CAN represent
+> - [IPNS][] names.
```
alanshaw, line 16: I'm not sure I understand the advantage of this...

Pin.md
```diff
@@ -0,0 +1,323 @@
...
+> - [did:key][] identifiers.
+> - Actors in [UCAN][] authorization.
+
+
+Documents can also be addressed in "specific state" by [CID][] _(which can be simply derived from CIDs of Shards it consists of, which will cover in more detail later)_ 
+
+
+### Document States
+
+Document can be in two logical states. Transitional state, here on referred as `Draft`, where it's in the process of transmittion _(e.g. in progress upload)_ where it has no `root`. Published document with a specific `root` is referred as `Edition`.
```
alanshaw, line 26: "Edition" - yes love it!

Pin.md
```diff
@@ -0,0 +1,323 @@
...
+
+#### Draft
+
+
+`Draft` represents in-flight document, usually while it's content is transmitted, which occurs prior to initial publish or between subsequent editions.
+
+
+It's `shard` field links to all the `Shard`s it is comprised of, which are [CAR][] encoded set of blocks.
+
+Every possible `Draft` state can be addressed by `CID` which can be computed by encoding it as DAG-CBOR node with `shards` sorted alphabetically (There's relevant prior art of [ZDAG hearder compression][] we colud borrow from)
```
alanshaw, line 58:
```suggestion
Every possible `Draft` state can be addressed by `CID` which can be computed by encoding it as DAG-CBOR node with `shards` sorted alphabetically (There's relevant prior art of [ZDAG header compression][] we colud borrow from)
```

I assume these are byte arrays so not really alphabetical but I get the idea.

Pin.md
```diff
@@ -0,0 +1,323 @@
...
+
+
+### Document States
+
+Document can be in two logical states. Transitional state, here on referred as `Draft`, where it's in the process of transmittion _(e.g. in progress upload)_ where it has no `root`. Published document with a specific `root` is referred as `Edition`.
+
+Following is an [IPLD schema][] definition for the `Document` object
+
+```ipldsch
+type Document union {
```
alanshaw, line 31: Do they not also have an ID?

Pin.md
```diff
@@ -0,0 +1,323 @@
...
+
+
+In order to reconcile concurrent publish operations we define total order (only) among published drafts as follows:
+
+1. Given drafts `D1` and `D2`, if all shards of `D1` are included in `D2` we say `D1 <- D2` (`D1` predates `D2`).
+2. Given drafts `D1` and `D2` where neither `D1` nor `D2` includes shards of the other `D1 <- D2` if:
+   1. Number of shards in `D2` is greater than in `D1`
+   2. Number of shards in `D2` is equal to number of shards in `D1` & `CID` of `D1 < D2`. 
+
+## Appliactions
```
alanshaw, line 144:
```suggestion
## Applications
```

alanshaw (commented):
This is super cool. I feel like we're almost there!

Pin.md
```diff
@@ -0,0 +1,252 @@
...
+-- In practice there will be Instruction set specific replica
+-- types
+type Change = Any
+```
+
+Semantics can more accuratly be captured with a help of generics, for that reason we present typescript definition below
+
+```ts
+type Replica<Change> {
+  prior?: Link<Replica<Change>>
```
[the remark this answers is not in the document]
Gozala: `Change` in this context meant to be `Append | Join`, more broadly speaking it should be closed set of operations.

Pin.md
```diff
@@ -0,0 +1,252 @@
...
+         type: "append",
+         shards: [
+           CID.parse('bagbaierauhgb4pxfuejvgufxxczjn2o7foetzrlxvnnvjm5pdas2y27v3cua')
+         ]
+       }
+     },
+     codec: CBOR,
+     hasher: sha256
+   })
+   ```
```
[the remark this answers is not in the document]
Gozala: I have an observable with bunch more examples that I plan on linking as well.

Pin.md
```diff
@@ -0,0 +1,350 @@
...
+1. Allow transient DAG representations, that is partially replicated DAGs or revisions of one with a traversable root. 
+1. Allow for an uncoordinated multiplayer DAG creation/transfer with specific convergence properties.
+
+### Motivation
+
+All content in [IPFS][] is represented by interlinked [blocks][IPLD Block] which form hash-linked DAGs. _(Every file in [IPFS][] is an [IPLD][] DAG under the hood.)_
+
+Many applications in the ecosystem have adopted [Content Addressable Archives (CAR)][CAR] as a transport format for (Sub)DAGs in settings where peer-to-peer replication is impractical due to network, device or other constraints. This approach proved effective in settings where CAR size limit is not a concern, however there are still many constrained environments (e.g. serverless stacks) where transferring large DAGs in single CAR is impractical or plain impossible.
+
+Here we propose a DAG replication protocol that overcomes above limitations by transporting large DAGs in multiple casually ordered network requests and/or sessions by:
```
expede, line 24:
```suggestion
Here we propose a DAG replication protocol that overcomes above limitations by transporting large DAGs in multiple causally ordered network requests and/or sessions by:
```

expede: (I'm assuming that you mean by causal relation, but could be wrong)

PR #5, opened 2022-09-16
w3-protocol spec

Gozala:
https://hackmd.io/@gozala/w3-protocol

### DIFF
--- a/w3-account.md
+++ b/w3-account.md
@@ -1,361 +1,400 @@
-# w3-accounts
+# Platform protocol
 
 ![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
 [![hackmd-github-sync-badge](https://hackmd.io/Zb7gjpLsQn2w3a3JUvnFcw/badge)](https://hackmd.io/Zb7gjpLsQn2w3a3JUvnFcw)
 
 ## Abstract
 
-Thinking about accounts in web2 terms introduces unfortunate limitations and seem to be a poor fit for User Controlled Authorization Network ([UCAN][]).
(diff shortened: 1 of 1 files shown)


hugomrdias:
should we leave comments here or hackmd ?

hugomrdias:
can we just merge this ? and create PRs to discuss changes with a smaller scope ?

w3-account.md
```diff
@@ -0,0 +1,1063 @@
...
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+### Representation
+
+Any valid [did:key][] identifier SHOULD represent a valid w3 account. Unknown accounts MAY be treated as an account with `0` balance/capacity and MAY be denied service.
+
+:::info
+All actors in w3 are also identified by [did:key][], but most [did:key]s will not correspond to accounts _(even if according to definition they technically are)_. Differnece between agent and account DIDs is purely in semantics and while same [did:key][] could be used for both, we enourage not to as two have to deal with very different threat models.
```
vasco-santos, line 38:
```suggestion
All actors in w3 are also identified by [did:key][], but most [did:key]s will not correspond to accounts _(even if according to definition they technically are)_. Difference between agent and account DIDs is purely in semantics and while same [did:key][] could be used for both, we encourage not to as two have to deal with very different threat models.
```

vasco-santos (commented):
Thanks for writing this up @Gozala  I read mostly for learning and don't have meaningful feedback at this point. Just putting all the pieces together in my head

PR #7, opened 2022-11-20
feat: draft did:dns and did:web spec

Gozala:


ucan-dns+web.md
```diff
@@ -0,0 +1,209 @@
...
+From which we would like to construct a delegation of the capability delegated in (1) issued by the the (rightmost) audience in (2).
+
+```
+did:key:zAlice -> did:dns:w3.storage -> did:key:zService -> did:key:zRotation -> did:key:zAli
+```
+
+To accomplish this we propose that:
+
+- To issue delegation from `did:key:zRotation` with a valid proof chain (2) proving that `did:key:zRotation` has been delegated all capabilities from `did:dns:w3.storage`.
+- Embed proof chain (1) in `fct` as a verifable evidence that `did:dns:w3.storage` has been delegated capabilities for `did:key:zAlice`.
```
expede, line 143: Interesting idea 💡 

This may open up some new avenues for a confused deputy because it relies on any possible UCAN that you learn about in the rest of the network. Arguably this similar to using the ambient authority at the root of a delegation chain, but involves anyone that delegates to the forwarding principal even if they're offline. Aside from phishing, I'll have to think more on specific attacks.

ucan-dns+web.md
```diff
@@ -0,0 +1,209 @@
...
+      iss: "did:key:zAlice",
+      aud: "did:dns:w3.storage",
+      exp: null,
+      att: [
+        {
+          with: "did:key:zAlice",
+          can: "*"
+        }
+      ]
+    },
```
expede, line 188: _If_ we decide to support pipelining, I wonder if we can put this in the `prf` field and just support this directly?

expede: (Since this is "evidence", which is another name for a proof!)

ucan-dns+web.md
```diff
@@ -0,0 +1,209 @@
...
+
+> There are other ways Alice could arrange delegation from `did:key:zAlice` to `did:key:zAli` with different tradeoffs. Evaluating them would be a good idea for a system designer, here we defining solution when no other option offers desired tradeoffs. -->
+
+To address described limitation we propose use of [`did:dns`] or [`did:web`] principals so that delegation from `did:key:zAlice` no loner is tied to a specific key. This would get us step further, but still we run into a problem, after key rotation delegation to `did:key:zAli` is no longer valid as the key that signed it is no longer the one did document resolves to.
+
+This specification describes solution to the second order problem by requiring that [`did:key`] that [`did:dns`] and [`did:web`] resolve to MUST assume an ambient authority over (pre-resolution) DID, which it MAY delegate to other principals through standard UCAN delegation.
```
expede, line 36: To clarify: this does not revoke the old key, correct? If so, then awesome 👍 

# If Key Revocation Intended

If you do want to revoke keys for cases like the key being compromised (by Mallory), this setup will not work. Mallory can simply not include the proof of rotation in her chain.

To actually _rotate_ a key without invalidating existing certificates, you need reference to some outside anchoring. This is a hard problem, and is the motivation behind many DID methods. A flow could look something like this:

1. Alice want to update her key from `did:key:zAlice1` to `did:key:zAlice2`
2. Alice registers a signed Merkle tree of existing UCAN delegations to a known source (a server, a DHT, etc)
3. Alice signals that she's retiring `did:key:zAlice1` in favour of `did:key:zAlice2`

This is not terribly unlike the current plan for eventually-consistent revocation. It _does_ mean that the DID either stops working under network partition (because it has to check an external system: DHT/server), or that Mallory can continue to use the retired key for an unbounded amount of time in the general case.

I don't totally dislike the eventually consistent model, since it matches user contract as revocation. Because it's gossip-based, you can update with web3.storage directly on invocation (since you're going to actually run the action anyways), so it's possibly viable in practice (though we'd need to write a bunch more about this in the spec)

ucan-dns+web.md
```diff
@@ -0,0 +1,209 @@
...
+  exp: null,
+  att: [],
+  prf: [{
+    iss: "did:dns:w3.storage",
+    aud: "did:key:zService",
+    exp: null
+    att: []
+  }]
+}
+```
```
expede, line 55: Because `att: []` means "delegate nothing", I think this doesn't work. Are you sure that you don't want delegation?


```json
{
  iss: "did:key:zService",
  aud: "did:key:zRotation",
  exp: null,
  att: [{
    iss: "did:dns:w3.storage",
    aud: "did:key:zService",
    exp: null
    att: []
  }],
  prf: [ fromOldKeyToService ]
}
```

expede: @matheus23 also came up with something for WNFS recently that may be interesting here (adapted to DID management)

```js
{
  iss: "did:dns:example",
  aud: "did:key:zService",
  exp: null,
  att: [{
    with: "did:dns:example",
    can: "did/rotate",
    nb: { to: "did:key:zNewKey" }
  }]
}
```

Here the UCAN only allows the DID document to be updated in a very specific way.

Per my earlier comment about revocation above, this would still require checking the list of valid public keys at read time. I don't think that you can get around that if you support revocation though.

expede: > Are you sure that you don't want delegation?

Another thing worth mentioning about the reasoning here: if the service can go rogue and change the key to anyone (including itself!), it's already equivalent to it having `*` rights to the delegated account. I don't think that you gain anything by making it a proof instead of a delegation.

ucan-dns+web.md
Gozala: To clarify: this does not revoke the old key, correct? If so, then awesome 👍

Correct!

> If you do want to revoke keys for cases like the key being compromised (by Mallory), this setup will not work. Mallory can simply not include the proof of rotation in her chain.

I think we may be misunderstanding each other here. Let me try to collaborate bit more and try to address your concert if I understand it correctly.

1. General idea is that at any time service is represented by chain of command with (at least) tree actors.
   - `did:dns:w3.storage = did:key:zW3 -> did:key:zService -> did:key:zRotation`
   - `did:key:zW3` is never meant to be rotated or used other than to create initial delegation to `did:key:zService`
      - It may be used in emergency of course in case `did:key:zService` did get compromised
   - `did:key:zService` is meant to be kept secure on hardware and only used to appoint `did:key:zRotation` key on regular schedule say once a month.
2. UCANs can essentially embed signed chain of command when issuing delegation:
     `did:dns:w3.storage = did:key:zW3 -> did:key:zService -> did:key:zRotation`
3. We're operating under optimistic assumption that `did:key:zService` just rotates keys and that they are not generally compromised.
4. For this reason we want delegation that embeds `did:dns:w3.storage = did:key:zW3 -> did:key:zService -> did:key:zPreRotation` to continue to work.
5. If however `did:key:zPreRotation` got compromised, we `did:key:zService` could revoke that last leg of delegation which will render embedded "signed chain of command" that includes it invalid.
    - we basically leverage UCAN revocations to revoke compromised keys, since we already operating with UCANs that makes a lot of sense.
6. So what happens when attacker who got hold of `did:key:zPreRotation` attempts to exploit and perform following delegation: `did:key:zAlice1 -> did:dns:w3.storage = did:key:zW3 -> did:key:zService -> did:key:zPreRotation -> did:key:zAttacker`.
    - UCAN validator starts by validating `did:key:zAlice1 -> did:dns:w3.storage` ✅
    - Then check `did:dns:w3.storage = did:key:zW3 -> did:key:zService` ✅ (unless resolving `did:dns:w3.storage` yields key different from `did:key:zW3`).
    - Then check `did:key:zService -> did:key:zPreRotation` ⛔️ - This got revoked by `did:key:zRotation`
7. In similar way we can rotate `did:key:zService` key with `did:key:zW3` and either revoke previous delegation or not.
8. If our `did:key:zW3` got compromised as well, we just update DNS record so that `did:dns:w3.storage = did:key:zW4`.
    - Now all the delegations will start failing, but that is desired in this scenario

expede: > Correct!

Okay! Amazing! Perhaps we should avoid the word "rotation" (`did:key:zRotation`), and maybe something more like "adding more keys" (or "append to the DID").

At this point it sounds a lot like "delegate all". I see that you've written a bunch more above. I'll run through that now 👀

expede: > did:key:zService is meant to be kept secure on hardware and only used to appoint did:key:zRotation key on regular schedule say once a month.

To double check my understanding:

You add a new key every month. So there's `zJan`, `zFeb`, `zMar`, etc. You then maybe retire (destroy) the old key out of band. As you say: under the assumption that they key is never compromised.

If that picture above is correct, I'm not sure that I understand the threat model. What does this get you?

ucan-dns+web.md
Gozala: > Because att: [] means "delegate nothing", I think this doesn't work. Are you sure that you don't want delegation?

I was kind of overloading it to mean redelegate everything it may be delegated to in the future.

> @matheus23 also came up with something for WNFS recently that may be interesting here (adapted to DID management)

How interesting, in some iteration of this document I had the same thing (although I called it `did/set` instead). However I have went back on that idea, because it felt too domain specific, I felt that it was bit confusing that ability to update DID key implied that I could delegate any capability delegated to that did.

I mean yes, but I wish UCAN had a more explicit thing than relying on DID specifics

> Note that in this case the DID itself is in the with field because you're updating the DID itself!

I have also considered following 

```js
{
  aud: "did:key:zRotation",
  iss: "did:key:zService",
  exp: null,
  att: [{ with: "did:dns:example", can: "*" }],
  prf: [{
    iss: "did:dns:w3.storage",
    aud: "did:key:zService",
    exp: null
    att: [{ with: "did:dns:example", can: "*" }],
  }]
}
```

But then concluded it was not right, because `{ with: "did:dns:example", can: "*" }` does not exactly imply `{ with: "did:key:zAlice", can: "*" } ` even if later was delegated to `"did:dns:example"`, or maybe it does but somehow it does not feel very clear to me.


I think this would be the clearest and most explicit that reads as `did:key:zService` can do anything with everything that `did:dns:w3.storage`. Just like `file/*` implies all current and future abilities under `file/` namespace so `with: "*"` could imply all the current and future resources principal has authority over.


```js
{
  iss: "did:dns:w3.storage",
  aud: "did:key:zService",
  exp: null,
  att: [{ with: "*", can: "*" }],
}
```

Yet another alternative would be to make `att` optional to imply an unattenuated delegation

```js
{
  iss: "did:dns:w3.storage",
  aud: "did:key:zService",
  exp: null,
}
```

ucan-dns+web.md
expede: Or maybe to expand on that question: what happens if `did:key:zJan` was compromised before it was retired. Won't that invalidate all of your downstream UCANs?

ucan-dns+web.md
Gozala: Oh, does this mean exactly what I was trying to express with various other forms ?

```js
{
  iss: "did:key:zService",
  aud: "did:key:zRotation",
  exp: null,
  att: [{
    with: "own://did:key:zOldKey/*",
    can: "*"
  }],
  prf: [ fromOldKeyToService ]
}
```

Or in other words does `did:key:zAlice` would be included there if delegation with `did:key:zAlice` to `did:key:zOldKey` exists ?

ucan-dns+web.md
Gozala: Please note that it is implied that:

1. Every delegation is still checked for revocation.
2. `did:dns:w3.storage` resolution still occurs and if it resolves to different key validation will fail.

It is true that you can incorporate any of the delegation you can find on the network, but unless principal it was delegated to redelegated to you it's not possible to utilize it.

ucan-dns+web.md
expede: Sorry, I feel like I'm coming off as a jerk today 😅 I guess I'm in hole-poking mode 🎈📌  I really DO like that this system only relies on UCAN token revocation. It's pretty cool, actually.

> At this point it sounds a lot like "delegate all".

It _is_ delegate all! Awesome 😄 

The main challenges as I see them:

1. An attacker could cause a lot of chaos by holding onto a key for a long time, and later misbehaving.
2. The chains will get pretty long (though this probably won't matter because we can memoize them by CID... so that's actually super awesome ✅)

Just to check, the chain looks like this:

```
serviceKey -> janKey -> febKey -> ...
```

...and not...

```
serviceKey 
   |
   +--> janKey
   |
   +--> febKey
   |
   +--> ...
```

...yes?

If so, is that because you want to keep the serviceKey offline except in emergencies, or because you don't want to maintain a DID document?

---

I have a really wacky alternate idea to get rid of the order dependence, and I _think_ fits your model... but I'm not sure that it 100% makes sense. Let me write it up.

ucan-dns+web.md
Gozala: I think we could, in fact I proposed that as option (2) in https://github.com/ucan-wg/spec/issues/130. I mostly avoided here as it would directly violate principal alignment requirement

ucan-dns+web.md
expede: > I mean yes, but I wish UCAN had a more explicit thing than relying on DID specifics

Tell me more! Can you describe more about the kind of properties that you wish it had?

Gozala: I pull relevant discussion into https://github.com/ucan-wg/spec/issues/131, probably better to continue this discussion there.

ucan-dns+web.md
expede: I'm catching up here, so bear with me.

I took a look at it, and the wacky idea is actually kind of interesting as a way to add new keys to a DID _offline_ over time if you control a root key by using accumulators. It doesn't work for your case, because you want to be able to target specific keys for revocation.

# Pipelining

I still have some concerns around attack vectors, but otherwise this fits capabilities really well! If this was an online ocap system, we'd have use a message forwarder. Doing this "statically" via UCAN is just the eventually consistent version.

This is great! As you mention, it would require a special kind of capability. Today the "delegate all" capability is _only_ for the resources directly owned by the referenced DID. There's nothing stopping us from making another ability that's something like `ucan/forward`.

In light of that: have you considered generating 100 years worth of monthly keys, delegating to them with this forwarder/pipeline capability, and storing offline until needed?

Gozala: > 
> You add a new key every month. So there's did:key:zJan, did:key:zFeb, did:key:zMar, etc. You then maybe retire (destroy) the old key out of band. As you say: under the assumption that the key is never compromised.
> 
> If that picture above is correct, I'm not sure that I understand the threat model. What does this get you?
>

Damage control

In the case we have to revoke the key, it will only affect UCANs that were issued by key during it's rotation. If we just rotate on compromise we would be affecting all the tokens every issued.

Gozala: No it actually looks like 

```
serviceKey 
   |
   +--> janKey
   |
   +--> febKey
   |
   +--> ...
```

Gozala: > In light of that: have you considered generating 100 years worth of monthly keys, delegating to them with this forwarder/pipeline capability, and storing offline until needed?

I have not. What would that exactly buy us over generating those keys as needed ? What I have considered instead is using multisgs for the `zService` principal

Gozala: > If so, is that because you want to keep the serviceKey offline except in emergencies, or because you don't want to maintain a DID document?

We want to keep top level key offline and reserve next key in chain of command only for appointing key in rotation so keep it "mostly" offline.

We also don't want to have to maintain DID document with growing number of keys, instead we could embed only relevant key delegations in UCANs that are affected by it. This makes things easier for both validators and ourselves, because all of the relevant data is in UCAN other than current DID key which we have no intention of rotating and could even hard code in our backend to avoid DID resolution there

ucan-dns+web.md
expede: I think the more that I think about it, the more comfortable I am with it 👍

ucan-dns+web.md
expede: > That is why if janKey is compromised we can revoke it but not febKey because that one was not compromised.

Ah, I'm glad that I asked! I thought that you were trying to keep the more powerful keys completely offline.

> What would that exactly buy us over generating those keys as needed ?

This was under the assumption that you were trying to keep the powerful keys totally offline. Never mind!

> We want to keep top level key offline and reserve next key in chain of command only for appointing key in rotation so keep it "mostly" offline.

Makes sense! I guess the "mostly offline" key is the high-value target because it's brought online semi-regularly.

> What would that exactly buy us over generating those keys as needed ?

Your powerful keys never come online. You can predelegate to them with `nbf` fields that match your rotation schedule. This means that your warm key can now completely cold. Because of the `nbf`, you don't have to be careful with the UCANs themselves, and can store them anywhere, in principle even in public.

> We also don't want to have to maintain DID document with growing number of keys

Actually, that's interesting. The wacky accumulator idea that I alluded to earlier basically does this because (if used the right way) it's a growable set at a stable identifier but still totally self-evident without needing to look up a central document. Cryptography is wild!

...I don't think that I need to go into it though, because the scheme you propose here potentially gets you more (specifically revocation).

Gozala: > This means that your warm key can now be completely cold. Because of the nbf, you don't have to be careful with the UCANs themselves, and can store them anywhere, in principle even in public.


Ah that’s awesome, now that I understand it it seems a great  idea

ucan-dns+web.md
```diff
@@ -0,0 +1,209 @@
...
+
+> There are other ways Alice could arrange delegation from `did:key:zAlice` to `did:key:zAli` with different tradeoffs. Evaluating them would be a good idea for a system designer, here we defining solution when no other option offers desired tradeoffs. -->
+
+To address described limitation we propose use of [`did:dns`] or [`did:web`] principals so that delegation from `did:key:zAlice` no loner is tied to a specific key. This would get us step further, but still we run into a problem, after key rotation delegation to `did:key:zAli` is no longer valid as the key that signed it is no longer the one did document resolves to.
+
+This specification describes solution to the second order problem by requiring that [`did:key`] that [`did:dns`] and [`did:web`] resolve to MUST assume an ambient authority over (pre-resolution) DID, which it MAY delegate to other principals through standard UCAN delegation.
+
+> Expample below illustrates `did:dns:w3.storage` delegating own resource to `did:key:zService`, which in turn redelegates it to `did:key:zRotation`.
```
hugomrdias, line 38:
```suggestion
> Example below illustrates `did:dns:w3.storage` delegating own resource to `did:key:zService`, which in turn redelegates it to `did:key:zRotation`.
```

ucan-dns+web.md
```diff
@@ -0,0 +1,209 @@
...
+  att: []
+}
+```
+
+It can also be interpreted as `did:dns:w3.storage` stating to be "also known as"
+`did:key:zService` allowing it to delegate whatever `did:dns:w3.storage` CAN.
+
+## Pipelining dellegations
+
+[UCAN][] specification does not offer a way for two delegation chains to be pipelined into one. In our described scenario we have two delegation chains:
```
hugomrdias, line 129:
```suggestion
[UCAN][] specification does not offer a way for two delegation chains to be pipelined into one. In our described scenario we have two delegation chains:
```

ucan-dns+web.md
```diff
@@ -0,0 +1,209 @@
...
+
+1. `did:key:zAlice -> did:dns:w3.storage`
+2. `did:dns:w3.storage -> did:key:zService -> did:key:zRotation`
+
+From which we would like to construct a delegation of the capability delegated in (1) issued by the the (rightmost) audience in (2).
```
hugomrdias, line 134:
```suggestion
From which we would like to construct a delegation of the capability delegated in (1) issued by the (rightmost) audience in (2).
```

ucan-dns+web.md
hugomrdias, line 143:
```suggestion
- Embed proof chain (1) in `fct` as a verifiable evidence that `did:dns:w3.storage` has been delegated capabilities for `did:key:zAlice`.
```

ucan-dns+web.md
```diff
@@ -0,0 +1,209 @@
...
+  aud: "did:key:zService",
+  exp: null
+  att: []
+}
+```
+
+It can also be interpreted as `did:dns:w3.storage` stating to be "also known as"
+`did:key:zService` allowing it to delegate whatever `did:dns:w3.storage` CAN.
+
+## Pipelining dellegations
```
hugomrdias, line 127:
```suggestion
## Pipelining delegations
```

hugomrdias (changes requested):
left some inline comments, i think the sections about full auth delegation and pipelining should be hoisted, there's examples that use those before they are defined. i was confused by that while reading.

PR #12, opened 2022-11-25
w3-accounts

Gozala:
Wrote a specification for explaining accounts and how they correlate with spaces

### DIFF
--- a/w3-account.md
+++ b/w3-account.md
@@ -218,7 +218,7 @@ However user may also add new delegations on one device and expect to have acces
  iss: "did:mailto:alice@web.mail",
  aud: "did:dns:web3.storage",
  att: [{
-   with: "did:mailto:web.mail",
+   with: "did:mailto:alice@web.mail",
    can: "access/claim"
  }],
  prf: [
--- a/w3-account.md
+++ b/w3-account.md
@@ -26,7 +26,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ## Motivation
 
-In web3.storage users MAY create number of user spaces simply by generating asymmetric key pair. They MAY be delegated capabilities for other user spaces. Managing these delegations and keypairs across multiple spaces, agents and devices can get complicated.
+In web3.storage users MAY create number of user spaces simply by generating asymmetric key pairs. User MAY also receive capability delegations from other user spaces. Managing these delegations and keypairs across multiple spaces, agents and devices can get complicated.
 
 To address this we propose a concept of an account, which is a [principal][] that can be delegated all relevant capabilities across various user spaces.
 
@@ -37,60 +37,27 @@ We also propose account authorization flow that would allowing allow authorized 
 
 ## Account
 
-User account is a [principal][] identified by [`did:mailto`][] (listed in a UCAN's `iss` or `aud` field).
+User account is a [principal][] identified by [`did:mailto`][] identifier.
 
-When user agent creates a new space, it MAY delegate full or subset of the capabilities to user account so they could be reclaimed by a user with another agent.
-
-> Example shows all capabilities to `did:key:zAlice` space been delegated to the user account `alice@web.mail` 
-
-```ts!
-{
-  iss: "did:key:zAlice",
-  aud: "did:mailto:alice@web.mail",
-  att: [{ with: "did:key:zAlice", can: "*" }],
-  exp: null,
-  sig: "..."
-}
-```
-
-## Delegations
-
-Agent MAY publish account delegations to a serivce so that they can be persisted and retrieved later with a different agent.
-
-> Invokes `access/delegate` asking web3.storage to record delegation from `did:key:zAlice` space to the `alice@web.mail` account.
-
- ```ts
-{
-  iss: "did:key:zAgent",
-  aud: "did:dns:web3.storage",
-  att: [{
-    with: "did:key:zAgent",
-    can: "access/delegate",
-    nb: {
-      access: Link<{
-        iss: "did:key:zAlice",
-        aud: "did:mailto:alice@web.mail",
-        att: [{ with: "did:key:zAlice", can: "*" }],
-        exp: null,
-        sig: "..."
-      }>
-     }
-    }
-  }],
-  sig: "..."
-}
-```
-
-:::warning
-[Recipient validation][] requires wrapping actual delegation into `access/delegate` invocation, but in the future we may find a way to remove this requirement.
+:::info
+It MUST be a DID identifier as opposed to `mailto:` URI to be a valid prinicipal in the UCAN protocol.
 :::
- 
+
+
+
 ## Authorization
 
-User MAY authorize an agent to represent their account by invoking `ucan/issue` capability with an audience where authorization is to be considered valid
+User MAY authorize an agent to represent their account by delegating capabilities to it. However since right now we have no way of creating or resolving [`did:mailto`][] documents, there is no supported way to issue such a delegation.
+
+:::info
+In the future we intend to address this by implementing support for [ucan mailto][] specification.
+:::
+
+To address this limitation service MUST provide `ucan/issue` capability, that user agent MAY invoke
+to get an authorization to act on behalf of the account.
 
 
-> Agent requests authorization to represent `alice@web.mail` with `did:key:zAgent` in the context of `web3.storage`
+> Exmaple illustrates authorization request to represent `alice@web.mail` with `did:key:zAgent` agent from `web3.storage`
 
 ```ts
 {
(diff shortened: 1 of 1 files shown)


w3-account.md
```diff
@@ -0,0 +1,224 @@
...
+      sig: "..."
+    }
+  ]
+}
+```
+
+
+# Limitations
+
+Using delegation from specific authority as an authorization proof limits it to the contexts where it is trusted. It is reasonable complomise when issued UCAN receipent is the same authority, but problematic in wider contexts.
```
vasco-santos, line 203:
```suggestion
Using delegation from specific authority as an authorization proof limits it to the contexts where it is trusted. It is reasonable compromise when issued UCAN receipent is the same authority, but problematic in wider contexts.
```

w3-account.md
```diff
@@ -0,0 +1,224 @@
...
+
+Long term we would like to upgrade this specification to [UCAN mailto][] so that authorazation may be verifiable and not require trusting third-party that performed [verification][].
```
vasco-santos, line 205:
```suggestion
Long term we would like to upgrade this specification to [UCAN mailto][] so that authorization may be verifiable and not require trusting third-party that performed [verification][].
```

vasco-santos (approved):
LGTM! Good writing @Gozala ❤️

w3-account.md
```diff
@@ -0,0 +1,224 @@
...
+  iss: "did:dns:web3.storage",
+  aud: "did:mailto:alice@web.mail",
+  att: [{
+    with: "did:dns:web3.storage",
+    can: "ucan/sign",
+    nb: { as: "did:key:zAgent" }
+  }],
+  exp: null
+  sig: "..."
+}
```
Gozala, line 136:
```js
{
iss: "did:dns:web3.storage",
aud: "...",
att: [{
  can: "send/message",
  with: "did:dns:web3.storage",
}],
prf: [{
  iss: "did:key:root",
  aud: "did:dns:web3.storage",
  att: [{
    with: "did:dns:web3.storage",
    can: "ucan/sign",
    nb: { as: "did:key:zAgent" }
  }],
  exp: null
  sig: "..."
}]
```

Gozala: @expede can you please provide feedback on this

w3-account.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+### issue `with`
+
+Resource MUST be a [`did:key`][] URI. It represents a public key that user wishes to use for signing UCANs issued by DID in the [`nb.as`][issue `as`] field.
+
+### issue `as`
+
+Account MUST be a [`did:mailto`][] principal. It is an account that user wishes to be represented by [`did:key`][] in the [`with`][issue `with`] field.
```
gobengo, line 113: See earlier comment. I don't think it's a good idea to have normative language that depends on a spec with no contents (`did:mailto`)

w3-account.md
```diff
@@ -0,0 +1,224 @@
...
+
+
+## Verification
+
+Service MUST perform out of bound user verification e.g. send a link to the mailbox of the account by clicking which user would authorize agent to represent the account.
```
gobengo, line 118:
```suggestion
Service MUST perform out of bound user verification. For example, the service could email a link to the mailbox of the account. Clicking the link in the email would authorize the agent to represent the account.
```

w3-account.md
```diff
@@ -0,0 +1,224 @@
...
+## Delegations
+
+Agent MAY publish account delegations to a serivce so that they can be persisted and retrieved later with a different agent.
+
+> Invokes `access/delegate` asking web3.storage to record delegation from `did:key:zAlice` space to the `alice@web.mail` account.
+
+ ```ts
+{
+  iss: "did:key:zAgent",
+  aud: "did:dns:web3.storage",
```
gobengo, line 65: I seem to recall somewhere you were deciding between whether to use `did:dns` or `did:web`. Where did that land? One tradeoff that comes to mind is that updates to `did:dns` may be more eventually consistent (because ttls), whereas `did:web` might give us more predictability about how fast we can expect updates to propagate.

gobengo:
I think it would be helpful to have something like a [sequence diagram](https://mermaid-js.github.io/mermaid/#/sequenceDiagram) explaining how the capabilities specified here would be used together to accomplish a use case.

w3-account.md
gobengo: @Gozala wrt `send/message`, that's new right? I notice that it is of the form `verb/noun` whereas most of the time our caps name convention is more like 'noun/verb'.

w3-account.md
```diff
@@ -0,0 +1,224 @@
...
+To address this we propose a concept of an account, which is a [principal][] that can be delegated all relevant capabilities across various user spaces.
+
+We also propose account authorization flow that would allowing allow authorized agent to act on behalf of the account [principal][].
+
+
+# Terminology
+
+## Account
+
+User account is a [principal][] identified by [`did:mailto`][] (listed in a UCAN's `iss` or `aud` field).
```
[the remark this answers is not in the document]
Gozala: > Why have this be did:mailto: and not mailto: ?

Because principals in UCAN spec are DIDs, if they were arbitrary URIs than using just `mailto:` URI would have made a lot more sense. Perhaps we can work towards allowing UCANs to use arbitrary URIs in principals, but I would pursue that separately and then update to just `mailto:` as opposed to getting blocked on that.

> If we do really want to define/use a did:mailto, how would did resolution work (e.g. [here is how did:dns resolution works](https://danubetech.github.io/did-method-dns/#resolve))

My plan had been do spec that out here https://github.com/ucan-wg/did-mailto, but I don't have bandwidth for that right now.

w3-account.md
Gozala: General idea is agnostic of choice between `did:dns` and `did:web` we could choose one that makes most sense or both. That said I think did:web is probably a going to be a better choice however

w3-account.md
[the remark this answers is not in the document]
Gozala: > how to create, read (resolve to did doc), update, and delete dids using that method.

In nutshell, resolution schema is you send them an email asking what's their DID document and they respond you back with did:key corresponding to their document

Gozala:
Refactored document a bit to align spec with user flow. Added a sequence diagram to illustrate the user flow

w3-account.md
Gozala: Hey @expede have you had a chance to consider this ? In our call, I believe you said you were ok with general idea, yet wanted to ponder more about actual representation

gobengo:
> Refactored document a bit to align spec with user flow. Added a sequence diagram to illustrate the user flow

@Gozala nice ty. any objection if at some point later I try to rewrite it as a mermaidjs sequence diagram so it's prettier? https://mermaid-js.github.io/mermaid/#/sequenceDiagram
* will be post-merge as its a stylistic improvement only

Gozala:
> @Gozala nice ty. any objection if at some point later I try to rewrite it as a mermaidjs sequence diagram so it's prettier? https://mermaid-js.github.io/mermaid/#/sequenceDiagram
> 
>     * will be post-merge as its a stylistic improvement only

No I'd be delighted if you do

w3-account.md
```diff
@@ -0,0 +1,269 @@
...
+
+> Exmaple illustrates authorization request to represent `alice@web.mail` with `did:key:zAgent` agent from `web3.storage`
+
+```ts
+{
+  iss: "did:key:zAgent",
+  aud: "did:dns:web3.storage",
+  att: [{
+    with: "did:key:zAgent",
+    can: "ucan/issue",
```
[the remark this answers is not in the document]
Gozala: It's more like `user/login`

w3-account.md
expede: Sorry for the post-merge feedback @Gozala !

> wrt send/message, that's new right? I notice that it is of the form verb/noun whereas most of the time our caps name convention is more like noun/verb

@gobengo I've been thinking about this in a UCAN "standard library" as `msg/send`, which cuts across email, text messaging, message passing, etc etc.


> ```
> iss: "did:dns:web3.storage", // <- Maybe this should be did:key:root here
> aud: "did:dns:web3.storage",
> ```

I continue to find `iss == aud` deeply strange. I agree with your comment "Maybe this should be did:key:root here".

Rather than granting the ability to sign "as" another DID, I think it's cleaner to make this a "forwarding edge", which follows much more the capabilities worldview. I _think_ that the end result is the same, so it's just naming.

```
Time 1: carol --[fwd_all]--> dan
Time 2: alice --[can: foo]--> bob --[can: foo]--> carol
Time 3: dan --[can: foo]--> frank
```

BECAUSE we can substitute a link:

```
alice --[can: foo]--> bob --[can: foo]--> carol
                                            |
                                        [fwd_all]
                                            |
                                            V
                                           dan --[can: foo]--> frank

End result: Frank can use Alice's foo
```

This is _kind of_ like "signing as" Carol, but it's only at the level of authority, not identity. _Alice_ saying "you can sign as Bob" is going to get really confusing unless you want to do DID management, which IMO should be at a DID layer. I think you can manage all of this cleanly at the layer of capabilities.

This is a powerful enough feature that has enough uses that I propose that we add this as a first-class feature to UCAN at some stage. In the meantime, I propose that we call this something other than "signing as".

Here's some delegations in this model:

```
w3s --[fwd_all]--> didJune
w3s --[fwd_all]--> didJuly
w3s --[fwd_all]--> didAug

didJune --[w3s's foo]--> alice

bob --[bob's bar]--> w3s
                      |
                   [fwd_all]
                      |
                      V
                   didJune --[bob's bar]--> carol
```

Here's that last one concretely (well, the `bob -> w3s -> june` portion just to keep it small _enough_ for an example)

```js
{
  iss: "did:key:june",
  aud: "did:key:carol",
  att: [{ can: "thing/doSomthing", with: "resource://example.com" }],
  sig: "...",
  prf: [
    {
      iss: "did:key:zBob",
      aud: "did:dns:web3.storage",
      att: [{
        with: "resource://exmaple.com",
        can: "thing/doSomething"
      }],
    exp: august2022,
    sig: "..."
  },
  {
      iss: "did:dns:web3.storage",
      aud: "did:key:june",
      att: [{
        with: "did:dns:web3.storage",
        can: "ucan/forward_all" // we can come up with a better label
      }],
      exp: november2022,
      sig: "..."
    }
  ]
}
```

PR #31, opened 2023-01-04
fix: did:mailto format per DID-core spec

### LINKED ISSUE #28: accounts spec uses dids with invalid syntax ('@' char)
Context
* reviewing 'accounts' spec https://github.com/web3-storage/specs/blob/main/w3-account.md
* the ['w3-space'](https://github.com/web3-storage/specs/blob/main/w3-space.md) spec essentially depends on this too 

Problem
* both the specs document using DIDs like `did:mailto:alice@web.mail`, and this string is not a valid did because it contains the '@' char in the `method-specific-id` section 
  * see [`did-core #did-syntax`](https://www.w3.org/TR/did-core/#did-syntax)
  * here is an issue where I confirmed with w3c ccg that unencoded `@` is not allowed - https://github.com/w3c/did-core/issues/829#issue-1336246541

Possible Solutions:
* we acknowledge that `@` is not allowed in DIDs, and that it must be percent-encoded (to `%40`) when in a did string
  * [ ] we update our specs to use this form of valid did string

Gozala:
Fix #28

I chose to not do percent encoding because:

1. I think it's going to be more error prone given that browsers do not escape `@` in URIs
2. [did:web](https://w3c-ccg.github.io/did-method-web/#example-creating-the-did-with-optional-path) uses this kind of encoding as opposed to arguably more intuitive `/` delimiter

### DIFF
--- a/w3-account.md
+++ b/w3-account.md
@@ -40,7 +40,7 @@ sequenceDiagram
   participant Email as 📬<br/><br/>alice@web.mail
 
   Agent ->> W3: access/authorize
-  Note right of Agent:🎟<br/>with: did:key:zAgent<br/>as: did:mailto:alice@web.mail
+  Note right of Agent:🎟<br/>with: did:key:zAgent<br/>as: did:mailto:web.mail:alice
   W3 ->> Email: ✉️ Verification email
   Email ->> W3: 🔗 Approve
   W3 -->> Agent: ./update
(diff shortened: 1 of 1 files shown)


gobengo (approved):
lgtm. I found myself wondering "what if email local-part is not something simple like 'alice', but instead has more unusual chars in it?". I assume we should expect percent-encoded entities in the 4th ':'-delimited segment of `did:mailto:domain:local-part`?

i.e.
```
const didMailto = `did:mailto:web.mail:alice-with-weird-chars`
const parts = didMailto.split(':')
const email = `${decodeUriComponent(parts[2])@${parts[1]}`
```

hugomrdias (approved):
LGTM

alanshaw (approved):

PR #33, opened 2023-01-09
💡Service as a DAG

Gozala:
You can see a [preview on hackmd](https://hackmd.io/@gozala/service-as-a-dag) (but please comment here not on hackmd).

Idea is not fully flashed out and mostly was inspired by schema limitations https://github.com/ipld/ipld/issues/263 and https://github.com/ipld/ipld/issues/262

I'm not 100% in love with the idea described, however there are few key discoveries that I think are worth considering / highlighting and incorporating into our work:

1. We should have a logical (DAG) model of our system.
   - Right now we have bunch of loose parts that connect, but no coherent picture. Having one would help ourselves and our customers
2. We should define a state machine for our model
   - We have vague idea how request goes through system pipeline, but again having an actual model would really help
      - It also would allow us to encode some invariant about which state transitions are impossible (e.g. from complete to pending).
3. Receipts need to provide complete trail of state changes
    - If customer has receipt for some request new receipts for that request should contain one user has
       - That helps with accountability if new state invalidates old one it should be captured
4. This one is bit abstract, but I think we naturally end up in situation where we have same set of operations (add, remove, list) across most namespaces.
    - Perhaps subconsciously we're already be desiging "service as a dag" ?
       - If so perhaps we should just make it explicit ? Would it lead to more simple to understand design ?
    - On the flip side graphql is leading you to think of service(s) as a graph yet it does not limit you to set of operations, instead it forces you to think of operations as queries and mutations
       - On the other hand "dag/post" described here is a generic "request" operation, we just formalize that target state starts off with whatever you've posted.

### DIFF
--- a/service-as-a-dag.md
+++ b/service-as-a-dag.md
@@ -107,10 +107,10 @@ type Operation union {
 # Task is similar to one in UCAN invocation draft spec, except it
 # has extra fields from invocation to make it self-contained as per
 # https://github.com/ucan-wg/invocation/issues/6
-type Task<Insruction> struct {
+type Task<Instruction> struct {
   with DID
   do String
-  input Insruction
+  input Instruction
   meta {String : Any} (implicit {})
   nnc optional String
   sig Varsig
@@ -119,9 +119,9 @@ type Task<Insruction> struct {
 
 # Roughly equivalent of the receipts from UCAN invocation spec
 # https://github.com/ucan-wg/invocation/blob/rough/README.md#9-receipt
-type Receipt<Insruction, State> struct {
+type Receipt<Instruction, State> struct {
   # Link to the "dag/put" task this is the receipt is for
-  task Task<Insruction>
+  task Task<Instruction>
   # Current state
   out State
   
@@ -146,7 +146,7 @@ type W3 struct {
   DID Space
 }
 
-# Spaces MAY have varios providers e.g storage provider, upload provider etc...
+# Spaces MAY have various providers e.g storage provider, upload provider etc...
 type Space struct {
   # Storege provider is map keyed by CARs stored. Users can send commands to
   # update or query it.


service-as-a-dag.md
```diff
@@ -0,0 +1,231 @@
...
+  # next set of entries.
+  Task<{ path IPLDPath limit optional Int }> "dag/select"
+} representation inline {
+  discriminantKey "do"
+}
+
+# Task is similar to one in UCAN invocation draft spec, except it
+# has extra fields from invocation to make it self-contained as per
+# https://github.com/ucan-wg/invocation/issues/6
+type Task<Insruction> struct {
```
gobengo, line 110:
```suggestion
type Task<Instruction> struct {
```

service-as-a-dag.md
```diff
@@ -0,0 +1,231 @@
...
+  with DID
+  do String
+  input Insruction
```
gobengo, line 113:
```suggestion
  input Instruction
```

service-as-a-dag.md
```diff
@@ -0,0 +1,231 @@
...
+  meta {String : Any} (implicit {})
+  nnc optional String
+  sig Varsig
+  prf [&UCAN]
+}
+
+# Roughly equivalent of the receipts from UCAN invocation spec
+# https://github.com/ucan-wg/invocation/blob/rough/README.md#9-receipt
+type Receipt<Insruction, State> struct {
```
gobengo, line 122:
```suggestion
type Receipt<Instruction, State> struct {
```

service-as-a-dag.md
```diff
@@ -0,0 +1,231 @@
...
+  # Link to the "dag/put" task this is the receipt is for
+  task Task<Insruction>
```
gobengo, line 124:
```suggestion
  task Task<Instruction>
```

service-as-a-dag.md
```diff
@@ -0,0 +1,231 @@
...
+With the above defined set of operation we can represent web3.storage as a DAG with a following schema. Users could execute above defined operations in order to mutate and query spaces they have access to
+
+```ipldsch
+# Entire system is modeled as map of user spaces keyed
+# by space DIDs
+type W3 struct {
+  DID Space
+}
+
+# Spaces MAY have varios providers e.g storage provider, upload provider etc...
```
gobengo, line 149:
```suggestion
# Spaces MAY have various providers e.g storage provider, upload provider etc...
```

Issue #45, opened 2023-02-01
update access/delegate spec to use `nb: {delegations: Dict<string,Link>}` instead of `nb: Dict<string,Link>`


gobengo:
because the former is easier to type using ucanto atm. `nb` doesn't take a schema, so can't use `Schema.dictionary()` on it

hugomrdias:
its `access/delegate` right ?

gobengo:
@hugomrdias Yes, thank you!

PR #47, opened 2023-02-01
feat!: access/delegate delegation set is at nb.delegations instead of nb root

gobengo:
Motivation:
* https://github.com/web3-storage/specs/issues/45

Note:
* this is backward incompatible, but afaict this protocol is pre-implementation, so I think that should be ok

### DIFF
--- a/w3-access.md
+++ b/w3-access.md
@@ -63,8 +63,10 @@ Agent CAN invoke `access/delegate` capability with arbitrary delegations which w
     {
       "with": "did:key:zAlice",
       "can": "access/delegate",
-      // Map of delegation links to be stored for their audiences.
-      "nb": { "bafy...prf1": { "/": "bafy...prf1" } }
+      "nb": {
+        // Map of delegation links to be stored for their audiences.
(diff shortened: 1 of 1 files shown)


Gozala (approved):

w3-access.md
```diff
@@ -93,7 +95,9 @@ Provider SHOULD deny service if DID in the `with` field has no storage provider.
 
 #### delegate `nb`
 
-Field is a set of delegations links represented as a JSON. Keys SHOULD be CID strings of the values to make encoding deterministic.
+##### delegate `nb.delegations`
+
+Field is a set of delegation links represented as JSON. Keys SHOULD be CID strings of the values to make encoding deterministic.
```
alanshaw, line 100: Why is this not an array? e.g. `[{ "/": "bafy...prf1" }]`

I assumed because the delegation could be inlined, but wording suggests otherwise.

gobengo: @alanshaw I think @Gozala originally used a dict to emulate a set and not have to require any sorting to get deterministic encoding.

PR #49, opened 2023-02-17
feat: non-sudo authorization

Gozala:
[🫣 preview](https://github.com/web3-storage/specs/blob/feat/non-sudo-auth/w3-session.md)


Gozala:
Closing in favor of #50

PR #51, opened 2023-03-01
feat: filecoin spec

vasco-santos:


w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
+# Filecoin Protocol
```
olizilla, line 1: `Filecoin Protocol` seems overloaded with the actual protocols that make up Filecoin.

one of Aggregate, Piece, Deal would be better (but not great)

vasco-santos: We have moved out of Aggregate because of historical namings confusion as well. I see this repo as protocols within web3.storage stack, and therefore Filecoin protocol looks fine. We can rename to `Filecoin deal Protocol` (or similar) if helpful

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+
+There are several roles in the authorization flow:
+
+| Name        | Description |
+| ----------- | ----------- |
+| Storage Provider | [Principal] identified by [`did:web`] identifier, representing a storage provider like w3up |
+| Authority   | [Principal] that represents service provider that executes invoked capabilities |
+| Verifier   | Component of the [authority] that performs UCAN validation |
+
+### Storage Provider
```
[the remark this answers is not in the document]
vasco-santos: I also felt weird considering Filecoin "Miners" are storage providers. But, we have also been calling ourselves storage provider before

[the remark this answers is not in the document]
olizilla: there was a big effort to get everyone to switch from "Filecoin miner" to "Storage Provider" from the Filecoin team. 

We cannot repurpose that term in this spec, as the intended audience will be Riba and any other teams that  provide the interface with Filecoin "Strorage Providers"... this doc needs to help disambiguate between us, the deal maker and the filecoin nodes that will hold the aggregations/pieces (aka storage providers)

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+
+# Protocol
+
+## Overview
+
+A Storage Provider is the entry point for user/application data into the web3. It will act on behalf of users to move data around into different storage points. One of the key storage points is Filecoin deals.
+
+## Authorization
+
+Storage Provider (web3.storage) must be given capabilities by a Filecoin broker Authority (Spade) to invoke capabilities to get its stored content into Filecoin deals.
```
dchoi27, line 55: I think this makes sense in a "push" world, but in a "pull" world, I think it can be somewhat simplified? In the push world, you need some sort of permission to invoke a method to get what you want. 

But in a pull world, you don't need anyone's permission to make data available via an endpoint (though you might want the opposite permission flow - have folks who are pulling from the endpoint need to have permission to access the data your endpoint is making available - so we might consider giving Spade a UCAN that delegates ability to query our endpoint). And then when Spade confirms that they're taking a set of CommPs off for dealmaking, they can send a receipt letting us know to no longer include them in the stack (this can probably be a UCAN as well if we delegate them some `dealengine/pop` method or something like that)

dchoi27: what they send us to confirm that the deal has been made is after that i guess. we'd at least need a receipt in UCAN form with the on-chain bits as proof (i don't think any actual "permissions" would necessarily need be invoked). but if we want it to be a permissioned flow, it could probably be similar to the above (we grant them permission to tell us a deal is made, they invoke this permission when it has, but they still should include the on-chain bits to make it so trust in spade isn't required to prove the deal was made)

alanshaw:
I was reading through this and noted down a slightly simplier version (IMO! haha)

https://hackmd.io/ZjfqlXaPRtO39rVzpOqy_A?view

I don't really like using "filecoin" everywhere, this is actually a service for submitting aggregates to another service that makes deals with Filecoin storage providers. It's pretty far removed.

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+sequenceDiagram
+    participant Authority as 🌐<br/><br/>did:web:spade.storage
+    participant Storage Provider as 🌐<br/><br/>did:web:web3.storage #32;
+
+    Storage Provider->>Authority: invoke `filecoin/add`
+    Note left of Authority: Prepare for deal
+    Authority-->>Storage Provider: receipt
+    Note left of Authority: Land deal in Filecoin
+    Authority->>Storage Provider: invoke `filecoin/deal`
+    Authority-->>Storage Provider: invoke `filecoin/fail`
```
Gozala, line 79: This part is not super clear to me. I'm assuming that spade wants to issuer receipt when request is acknowledged, yet succeed or fail it later.

If so [UCAN invocation spec has `.fx` field on the receipts](https://github.com/ucan-wg/invocation/#7-effect) to cover for this. Specifically it could point to another task in the workflow which can be then used to checked.

In other words instead of having `filecoin/deal` and `filecoin/fail` capabilities to report the status, I recommend `receipt` have a `fx.join` field pointing to e.g. `filecoin/devise` task which will eventually get receipt with either `ok` or `error`.

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+## Capabilities
+
+This section describes the capabilities that form the w3 Filecoin protocol, along with details relevant to invoking capabilities with a service provider.
+
+### `filecoin/add`
```
Gozala, line 86: Nit: I feel like `filecoin/offer`, `filecoin/submit` or perhaps `filecoin/post` would better convey the fact that it offers some data while it's up to the recipient to accept or a deny it.

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+> `did:web:web3.storage` invokes capability from `did:web:spade.storage`
+
+```json
+{
+  "iss": "did:web:web3.storage",
+  "aud": "did:web:spade.storage",
+  "att": [{
+    "with": "did:web:spade.storage",
+    "can": "filecoin/add",
+    "nb": : {
```
Gozala, line 111:
```suggestion
    "nb": {
```

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+      "aggregateId": "id...",
```
Gozala, line 112: What is this id derived from ?

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+          "link": "bag...",
+          "size": 1234,
+          "md5": "md5...",
+          "commP": "commP...",
+          "url": "https://..."
+        }
+      ]
+    }
+  }],
+  "prof": [
```
Gozala, line 124:
```suggestion
  "prf": [
```

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+A Filecoin broker authority MAY invoke deal capability when one aggregate lands into a Filecoin deal. It should include the CID of the request CAR cid from `filecoin/add` invocation.
+
+```json
+{
+  "iss": "did:web:spade.storage",
+  "aud": "did:web:spade.storage",
+  "att": [{
+    "with": "did:web:spade.storage",
+    "can": "filecoin/deal",
+    "nb": : {
```
Gozala, line 146:
```suggestion
    "nb": {
```

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+    "exp": null,
+    "sig": "..."
+  ],
+  "sig": "..."
+}
+```
+
+### `filecoin/deal`
+
+A Filecoin broker authority MAY invoke deal capability when one aggregate lands into a Filecoin deal. It should include the CID of the request CAR cid from `filecoin/add` invocation.
```
Gozala, line 137: As mentioned earlier I think it's best to leverage effects and have general interface for subscribing for the updates. But assuming we choose to not use them here, I would suggest using a different namespace on the web3.storage side because:

1. I would like to be able to say `web3.storage` implements `filecoin/*` protocol which can not be true if some capabilities are part of the other end of the pipeline. In other words I think it's good idea to have `/add` be in the different namespace from `/deal` and `/fail`

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+    }
+  }],
+  "prof": [],
+  "sig": "..."
+}
+```
+
+### `filecoin/fail`
+
+A Filecoin broker authority MAY invoke deal capability when one aggregate fails to land into a Filecoin deal. It should include the CID of the request CAR cid from `filecoin/add` invocation.
```
Gozala, line 159: I assume there are no partial failures per aggregate. If so I think having one invocation with `result` type telling if it's success or failure is a better design

w3-filecoin.md
```diff
@@ -0,0 +1,187 @@
...
+
+```json
+{
+  "iss": "did:web:spade.storage",
+  "aud": "did:web:spade.storage",
+  "att": [{
+    "with": "did:web:spade.storage",
+    "can": "filecoin/deal",
+    "nb": : {
```
Gozala, line 168:
```suggestion
    "nb": {
```

w3-filecoin.md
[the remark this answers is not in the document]
vasco-santos: By this

> Specifically Spade can register web3.storage as a customer and allow allow invocations signed by it.

I understand by this that Spade would issue proofs of delegation `aggregate/*` that web3.storage could use

Therefore, looks like this was what is mentioned. But I will flip the clauses to make it easy to read

w3-filecoin.md
vasco-santos: Thanks, this makes total sense. I think there are two angler here:

- Confirmation receipt, which totally makes sense by your suggestion and I will add it here!
- Getting state of a deal over time. Like today we have `api.web3.storage/status/:cid` where users can check details about where given CID underlying content is stored, this is a way to give feature parity. In other words, be able to ask Spade about deal information for a given aggregate identifier.

vasco-santos: Also note that we can't store this information on web3.storage side given Broker will be responsible to renew deals over time and update this information

vasco-santos: One question @Gozala about effects. Would then `filecoin/devise` be with `issuer` and `audience` both Spade in this case?

w3-filecoin.md
vasco-santos: I totally agree! Moving now to aggregate even makes more sense to me

w3-filecoin.md
vasco-santos: Yes, everything fails. We get a list of chunks that made it fail

Gozala (changes requested):
Submitted a feedback inline, tldr;

1. I think we should frame it as a following steps
   1. Storefront offers broker an aggregate
   2. Broker queues the offer and returns effect that when performed will review the offer
   3. Broker reviews an offer and issues receipt for either accepting (success) or denying (failure) the offer.
      - Once offer is accepted broker takes care of arranging & renewing deals
   4. Storefront can query state of the aggregate deals using `aggregate/get`

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+
+# Protocol
+
+## Overview
+
+A Storefront is the entry point for user/application data into web3. It will act on behalf of users to move data around into different storage points. One of the key storage presences may be Filecoin Storage Providers.
+
+### Authorization
+
+Broker MUST have an authorization mechanism for allowed Storefront principals (e.g. web3.storage). Either by out-of-bound exchange of information or through a well defined API. In other words, broker can authorize invocations from `did:web:web3.storage` by validating signature from did. This way, it allows web3.storage to rotate keys without having too coordinate that with the broker.
```
alanshaw, line 59:
```suggestion
Broker MUST have an authorization mechanism for allowed Storefront principals (e.g. web3.storage). Either by out-of-bound exchange of information or through a well defined API. In other words, broker can authorize invocations from `did:web:web3.storage` by validating signature from did. This way, it allows web3.storage to rotate keys without having to coordinate with the broker.
```

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+
+```mermaid
+sequenceDiagram
+    participant Authority as 🌐<br/><br/>did:web:spade.storage
+    participant Storefront as 🌐<br/><br/>did:web:web3.storage
+
+    Storefront->>Authority: invoke `aggregate/offer`
+    Note left of Storefront: Queueing offer
+    Authority-->>Storefront: receipt issued
+```
```
alanshaw, line 88: This one is confusing because it's very similar to the one above, but the parties are swapped around. Do we need this diagram? Can we just put `Authority-->>Storefront: receipt issued` in the one above?

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+
+### Broker reviews and handles the offer
+
+When a broker pops the offer from the queue, the offer details MUST be retrievable. With the offer details, the broker MAY interact with available Filecoin Storage Providers, in order to establish a previously determined number of deals. Depending on storage providers availability, as well as the content present in the offer, the aggregate MAY be handled or not. A receipt is created to proof the transition of `aggregate/offer` state from `queued` into `accepted` or `denied`.
```
alanshaw, line 92: > When a broker pops the offer from the queue, the offer details MUST be retrievable.

"pops" suggests a FILO structure, i.e. not a queue, a stack.

I'm not sure I understand what this means - might need re-wording.

Is the offer status not availabile for querying until it reaches the front of the queue? Oh, you mean the broker should be able to retrieve the details of the aggregate from us?

> MAY interact with available Filecoin Storage Providers

Why is this may? It _will_ interact with SPs to facilitate a deal.

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+
+```mermaid
+sequenceDiagram
+    participant Authority as 🌐<br/><br/>did:web:spade.storage
+    participant Storefront as 🌐<br/><br/>did:web:web3.storage
+
+    Note left of Storefront: Review and handle offer async
+    Authority-->>Storefront: receipt issued
+```
```
alanshaw, line 101: Can we keep the side the participants are on static i.e. keep web3.storage on the left?

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+
+Once offer is accepted broker takes care of arranging and renewing deals.
+
+### Storefront can query state of the aggregate deals
+
+Storefront users MAY want to check details about deals from the content they previously stored. These deals will change over time as they get renewed. Therefore, Storefront should invoke `aggregate/get` capability to gather information about given aggregate identifier.
```
alanshaw, line 107: For me, `aggregate/get` suggests we're retrieving the aggregate data, which is not the case, we're retrieving information _about_ the deals the aggregate is included in.

`aggregate/info` `aggregate/status` or `aggregate/deals` (I'm guessing this one was disapproved already)

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+    "with": "did:web:web3.storage",
+    "can": "aggregate/get",
+    "nb": {
+      "commP": "commP...",
+    }
+  }],
+  "prf": [],
+  "sig": "..."
+}
+```
```
alanshaw, line 204: What does the return (receipt) for this invocation look like?

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+
+### `offer/review`
+
+When a broker receives an `aggregate/offer` invocation from a Storefront Principal, an [Effect](https://github.com/ucan-wg/invocation/#7-effect) for this submission is created with join task to be performed asynchronously.
```
alanshaw, line 208: Not sure I understand this part - is this the delegation that `bafy...dequeue` referrs to in `fx.join`?

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+
+Once this invocation is executed, a receipt is generated with the status of the task updated. Accepted aggregate receipt will provide aggregate status info:
+
+```json
+{
+  "ran": "bafy...review",
+  "out": {
+    "ok": {
+       "status": "accepted",
+       "link": "bafy...aggregate"
```
alanshaw, line 234: Why do we need this field?

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+```
+
+If offered aggregate is invalid, details on failing commPs are also reported:
+
+```json
+{
+  "ran": "bafy...invocation",
+  "out": {
+    "error": {
+      "status": "denied",
```
alanshaw, line 253:
```suggestion
      "status": "rejected",
```

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+}
+```
+
+### Schema
+
+```ipldsch
+type Aggregate =
+  { "status": "queued", link: Link } |
+  { "status": "accepted", link: Link } |
+  { "status": "denied", link: Link }
```
alanshaw, line 275:
```suggestion
  { "status": "rejected", link: Link }
```

alanshaw (changes requested):

w3-aggregation.md
vasco-santos: I aligned with the role that was acting in each side of the flow. But no strong opinion, so moving to that

w3-aggregation.md
```diff
@@ -0,0 +1,337 @@
...
+{
+  "iss": "did:web:web3.storage",
+  "aud": "did:web:spade.storage",
+  "att": [{
+    "with": "did:web:web3.storage",
+    "can": "aggregate/offer",
+    "nb": {
+      "offer": {
+        "link": "bagy...aggregate",
+        "size": 110101,
```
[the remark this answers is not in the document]
vasco-santos: data it contains, size of all the CARs together. We can't comment JSON, but I can add a note above

w3-aggregation.md
vasco-santos: Added https://github.com/web3-storage/specs/pull/51/commits/dae9ff404015b8eecfe07e4c9384937444b9494d

It includes array of deals information in that moment. probably we should also add a timestamp there to say this was what was there at this point based on what we talked about yesterday?

w3-aggregation.md
vasco-santos: yes!

w3-aggregation.md
vasco-santos: Reworded in https://github.com/web3-storage/specs/pull/51/commits/91e928080484202966556f87d020404fb31035fe

> Oh, you mean the broker should be able to retrieve the details of the aggregate from us?

yes!

w3-aggregation.md
```diff
@@ -0,0 +1,293 @@
...
+### `aggregate/submit`
+
+A Storefront principal can invoke a capabilty to submit an aggregate ready for deals.
+
+> `did:web:web3.storage` invokes capability from `did:web:spade.storage`
+
+```json
+{
+  "iss": "did:web:web3.storage",
+  "aud": "did:web:spade.storage",
```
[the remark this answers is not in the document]
ribasushi: I am personally not excited ( because I hate `deal` as a concept in general, and have phased out that language within spade entirely ). However @dkkapur is running the client-facing show, so deferring to him :)

vasco-santos:
@Gozala @alanshaw iterated on this based on previous convos. Essentially, handled 2 out of the 3 key take aways we had:
- not have aggregate commP anywhere in the protocol
- drop CARs for encoding CAR pointers and use CBOR block instead

did not added new state that is between queued and accepted because we currently do not have any new info to add there and talking with Alan we decided maybe not to worry about this for MVP

w3-aggregation.md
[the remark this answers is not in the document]
Gozala: What the term used by spade ?

w3-aggregation.md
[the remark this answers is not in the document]
Gozala: I think it would be go a good idea to write down an IPLD Schema not just examples.

w3-aggregation.md
```diff
@@ -0,0 +1,359 @@
...
+```json
+[
+  {
+    "link": "bag...",
+    "size": 110101,
+    "commP": "commP...",
+    "src": ["https://.../bag(...).car"]
+  } 
+]
+```
```
Gozala, line 151: I was under impression that in the last call we decided we don't want to link to the CAR containing these entries but rather have `offer` field contain these entries inline instead

w3-aggregation.md
```diff
@@ -0,0 +1,359 @@
...
+
+## Capabilities
+
+This section describes the capabilities that form the w3 aggregation protocol, along with the details relevant for invoking capabilities with a service provider.
+
+In this document, we will be looking at `spade.storage` as an implementer of the `aggregate/*` and `offer/*` protocol.
+
+### `aggregate/offer`
+
+A Storefront principal can invoke a capabilty to submit an aggregate ready for offers.
```
Gozala, line 116:
```suggestion
A Storefront principal can invoke a capabilty to submit an aggregate ready for offers.

```iplsch
type AggregateOffer union {
  Aggregate  "aggregate/offer"
} representation inline {
  discriminantKey "can"
}

type Offer struct {
  with   StorefrontDID
  offer  { CARLink: OfferDetails }
}

type struct OfferDetails {
    size        Int
    commP       Bytes
    src         [URL]
}

type CARLink string
```

I think adding a schema here would be very helpful. Note that I have inlined details as per our last call. Also instead of a list I've used map keyed by CAR cid, which I think is better as it's more deterministic (agnostic of order).

w3-aggregation.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+Once this invocation is executed, a receipt is generated with the aggregate information:
+
+```json
+{
+  "ran": "bafy...get",
+  "out": {
+    "ok": {
+      "deals": [
```
Gozala, line 203: I would personally use map keyed by `dealId` so that it's order agnostic.

w3-aggregation.md
```diff
@@ -0,0 +1,359 @@
...
+    "nb": {
+      "link": "bafy...aggregate",
+    }
+  }],
+  "prf": [],
+  "sig": "..."
+}
+```
+
+Once this invocation is executed, a receipt is generated with the aggregate information:
```
Gozala, line 196: I think we need to consider two things here:

1. How does cashing affects this ? Ideally invocations are idempotent meaning same instruction should get same result (even if state has changed in the meantime) and `nnc` supposed to be utilized to cache miss. If we do not want to deal with caching we should specify additional field like current time and request that it be set to e.g. latest drand or a wall clock, that way out dated responses are less surprising, if you request state for old state that has not been computed prior you get an error saying it can not be computed, if we have it cached than response is for that time frame.

1. We should specify what happens if you request deal info for the aggregate that is in `queued` state.

w3-aggregation.md
Gozala: Since in last call we decided offer reviews is not really a thing and instead spade just attempts to arrange them and either succeeds or fails I don't think previously suggested `offer/review` still makes sense, probably should be called `offer/arrange` instead ?

w3-aggregation.md
Gozala: Yeah for convenience. If we don't have pointer to the aggregate you will not know for which aggregate this receipt is. You would have to first resolve invocation block from `ran` field, and find out the aggregate from it's `nb.link`. 

All of this can be made even more complicated if you don't have necessary blocks locally. Including the link here means you'll know all from the receipt without having to do any resolution.

w3-aggregation.md
```diff
@@ -0,0 +1,359 @@
...
+```
+
+If offered aggregate is invalid, details on failing commPs are also reported:
+
+```json
+{
+  "ran": "bafy...invocation",
+  "out": {
+    "error": {
+      "status": "rejected",
```
Gozala, line 275: I wonder if we need `status` field at all since `ok` / `error` kind of tell us everything we need to know already.

w3-aggregation.md
```diff
@@ -0,0 +1,359 @@
...
+      "link": "bafy...aggregate",
+      "cause": [{
```
Gozala, line 277: I think I'd use map keyed by car CID here as well.

w3-aggregation.md
```diff
@@ -0,0 +1,359 @@
...
+}
+```
+
+### Schema
+
+```ipldsch
+type Aggregate =
+  { "status": "queued", link: Link } |
+  { "status": "accepted", link: Link } |
+  { "status": "rejected", link: Link }
```
Gozala, line 298: This is not a valid IPLD schema syntax, following is:

```suggestion
type Aggregate union {
  | AggregateInfo "queued"
  | AggregateInfo "status"
  | AggregateInfo "rejected"
} representation inline {
  discriminantKey "status"
}

type AggregateInfo struct {
  link: Link
}
```

Gozala: That said I think using [keyed representation](https://ipld.io/docs/schemas/features/representation-strategies/#union-keyed-representation) is more idiomatic so we may want to use following instead:

```suggestion
type Aggregate union {
  | Link "queued"
  | Link "status"
  | Link "rejected"
} representation keyed
```

Gozala (approved):
Provided some suggestions but overall I think it looks good to me. My only concern is following, for which I have no good solution:

In OCAP / UCAN model only resource owner or it's delegate should be able to mutate state, that is to say if storefront is an owner of `aggregate/*` it should be the one doing state changes and broker should have no ability to do so without delegation. However `aggregate/get` seems to query state of the resource that is changed by broker, which does not seem good.

More OCAP-y approach would allow store front to mutate `aggregate/*` namespace make broker in charge of another resource e.g. `offer/*` namespace. Idiomatic approach would be to make `offer/arrange` either result in error or a success that delegates capability to query the deals for the submitted aggregate.

Idiomatic approach does have a downside that to query deals you need to first get delegation from the `offer/arrange` result and manage that state. On the other hand it does answer some of the questions I've listed inline like what's the response if you ask for state while it's in the queue ? You simply won't be able to because you won't have delegation to do so.

Gozala (changes requested):
I meant to request changes instead of approving.

w3-aggregation.md
```diff
@@ -0,0 +1,359 @@
...
+
+### Broker queues the offer
+
+Once broker successfully gets an offer, the offer gets queued for review. A receipt is created to proof the transition of `aggregate/offer` state from `null` into `queued`.
+
+This receipt MUST contain a follow up task in the (`.fx.join` field) that is run when submitted request is processed. It MAY succeed (if aggregate was accepted) or fail (if aggregated was determined to be invalid). The result of the subsequent task CAN be looked up using its receipt.
+
+### Broker reviews and handles the offer
+
+When a broker dequeues the offer from the queue, the offer details MUST be retrievable. With the offer details, the broker will interact with available Filecoin Storage Providers, in order to establish a previously determined number of deals. Depending on storage providers availability, as well as the content present in the offer, the aggregate MAY be handled or not. A receipt is created to proof the transition of `aggregate/offer` state from `queued` into `accepted` or `denied`.
```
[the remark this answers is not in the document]
vasco-santos: > It's not clear to me what this tries to convey

rephrased

> This is also confusing, because I don't think we mentioned how this number is determined.

added context that it is "out of band". It is something agreed with Spade out of the protocol.

w3-aggregation.md
```diff
@@ -0,0 +1,359 @@
...
+```mermaid
+sequenceDiagram
+    participant Storefront as 🌐<br/><br/>did:web:web3.storage
+    participant Authority as 🌐<br/><br/>did:web:spade.storage
+
+    Note right of Storefront: Review and handle offer async
+    Authority-->>Storefront: receipt issued
+```
+
+Once offer is accepted broker takes care of arranging and renewing deals.
```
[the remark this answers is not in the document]
vasco-santos: > we should rephrase this so it does not leave a wrong impression.

Good point, by accepted I mean `accepted` state. But I will make it more clear

w3-aggregation.md
vasco-santos: everything is a bit overloaded terms. It depends on what we consider an aggregate. If we do not assemble one (is just metadata about one in the new world, given Spade also not assemble it...), I would not expect get to get the assembled thing.

We can align on naming when we can sync and update if needed

w3-aggregation.md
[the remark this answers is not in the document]
vasco-santos: the schema is below and also includes the offer type

w3-aggregation.md
vasco-santos: you mean `commP`?

w3-aggregation.md
Gozala: That is reasonable, lets just add sentence to state that.

w3-aggregation.md
Gozala: No I meant car CID, but maybe that is the mistake. Probably should be keyed by whatever we'll key the input to the `aggregate/offer`

w3-aggregation.md
Gozala: I agree! Lets separate general review from the naming which we can do in a separately in sync.

w3-aggregation.md
```diff
@@ -0,0 +1,380 @@
...
+
+```json
+{
+  "iss": "did:web:web3.storage",
+  "aud": "did:web:spade.storage",
+  "att": [{
+    "with": "did:web:web3.storage",
+    "can": "aggregate/get",
+    "nb": {
+      "link": "bafy...aggregate",
```
Gozala, line 211: We no longer have CID for the aggregate since we started inlining, and I'm not sure if deriving CID from the `nb.offer` to ask it's status is still making sense. Yet I do like deterministic identifier as opposed to made made up ones.

w3-aggregation.md
```diff
@@ -0,0 +1,380 @@
...
+  "fx": {
+    "fork": []
+  },
+  "meta": {},
+  "iss": "did:web:spade.storage",
+  "prf": []
+}
+```
+
+### `offer/arrange`
```
Gozala, line 250: I think moving this section after `aggregate/offer` might make it easier to follow the flow, although interleaving `aggregate/` and `offer/` could make it confusing in a different way.

Perhaps a sentence and a link to this section after `aggregate/offer` would be a good compromise.

w3-aggregation.md
Gozala: I mentioned it elsewhere, but inline here is probably going to make more sense.

In terms of capabilities it does seem wrong that broker is able to mutate `did:web:web3.storage` resource without been delegated capability to do so. And that is kind of what happens as a side effect of `offer/arrange`.

> Although perhaps we could make an argument that by delegating `aggregate/offer` you do delegate ability to also update it's state.

I think it would be more cleaner if instead `offer/arrange` produced a receipt which in success case would include delegation to `offer/*` capabilities to the storefront which would allow it then to query for the state using `offer/get` capability. With such arrangement it would make sense that spade can update own state and that storefront will be able to query it because it was delegated capabilities to do so.

As an added benefit it would also make it impossible to query for the status until aggregate was accepted, because storefront simple will not have capability to do so.

The downside of going this route is that storefront would need to maintain delegations for the accepted aggregates to query the state. However we could probably find a way around this by adding additional `nb.storefront` field into `offer/*` capabilities so that one delegation could cover all of them.

Gozala (commented):
Provided more feedback and suggestions.

w3-aggregation.md
vasco-santos: This `bafy...aggregate` in here would be the CID of the CBOR block where we inline offer details. Maybe the issue is using the term aggregate, should we set `bafy.....offer` instead?

vasco-santos: ok wait, by your suggestion above now looks like we are not aligned in this. So, from my understanding of other meeting we would like to have all CARs to be part of the offer encoded in a single CBOR block, instead of within a CAR file with a block per entry.

With the above in mind, I would expect us to have a list with metadata of CARs (commP, cid, size), which would be encoded with DAG CBOR. Our service has an endpoint where this information can be fetched by the CID of the DAG CBOR.

In protocol: 

```
"nb": {
      "offer": {
        "link": "bafy...aggregate",
        "size": 110101,
        "src": ["https://w3s.link/ipfs/bafy...aggregate"]
      }
    }
```

`bafy...aggregate` would be the CID of the block which has information of all the CARs that are part of the offer. (perhaps we should call it `bafy...offer` to avoid confusion)


From [comment](https://github.com/web3-storage/specs/pull/51#discussion_r1163667443) it looks like @Gozala would be expected different approach where we just inline directly in `nb` the offer content (map with CID of each CAR to be included). The problem with that is that we do not have an identifier for the offer. We can derive it by calculating a CID for `nb`, but in my opinion that adds extra complexity. Moreover, offer content might be 10.000s of CAR files, which I would like to avoid bloating invocations size that go into the stream.

w3-aggregation.md
vasco-santos: https://github.com/web3-storage/specs/pull/51/commits/ec12c091359ed7c186aa97cfb1aba961e714cbc5

w3-aggregation.md
vasco-santos: Adding what I added in async convo:

in short, question here is how to represent offer sent from w3filecoin pipeline to spade-proxy in the UCAN protocol.
Offer contains the information for an aggregate of CAR files to be created. i.e a list of metadata of each CAR to be included (cid, commP, size, sourceUrl)

There are two alternatives
Encode offer with DAG CBOR, and provide in the UCAN capability invocation the cid for the offer, together with offer total size and sourceUrl
Inline in UCAN capability invocation metadata of all CAR files to be included

Expanding a bit, option 1 would look like:
"nb": {
  "offer": {
    "link": "bafy...offer",
    "size": 110101,
    "src": ["https://filecoin.web3.storage/bafy...offer"]
    }
}

While option 2 would look like

"nb": {
  "offer": {
      "bafy...car0": {
	       "size": 110101,
	        "commP": { "/": { "bytes": "7aEDQ...ZQQ" } },
	        "src": ["https://w3s.link/ipfs/bag...car0"]
      },
      "bafy...car1": {
	       "size": 110101,
	        "commP": { "/": { "bytes": "7aEDQ...ZQQ" } },
	        "src": ["https://w3s.link/ipfs/bag...car1"]
      }
      ...
  }
}

In option 1, we would get block that would have format of option 2.

With option 2 we can derive a CID as an identifier, by getting a CID for the nb object. While in option 1 the identifier CID is explicit

w3-aggregation.md
```diff
@@ -0,0 +1,396 @@
...
+    "commitmentProof": { "/": "commitment...car0" }, /* COMMP CID */
+    "src": ["https://w3s.link/ipfs/bag...file0"]
+  },
+  {
+    /* ... */
+  }
+]
+```
+
+Each entry of the decoded offers block, has all the necessary information for a Storage Provider to fetch and store a CAR file. The `link` field has the CAR CID, while the `commitmentProof` field has the required `proof` bytes by Storage Providers (for example, `commP`). The `src` field of each piece MUST be set to a (alphabetically sorted) list of URLs from which it can be fetched. The `size` field MUST be set to the byte size of the CAR file.
```
alanshaw, line 181: Do we want to specify that the `src` field is optional (allowing it to be provided when the deal is signed)?

w3-aggregation.md
```diff
@@ -0,0 +1,396 @@
...
+
+Broker MUST issue a signed receipt to acknowledge the received request. Issued receipt MUST contain an [effect](https://github.com/ucan-wg/invocation/#7-effect) with a subsequent task (`.fx.join` field) that is run when submitted aggregate is processed and either succeeds (implying that aggregate was accepted and deals will be arranged) or fail (with `error` describing a problem with an aggregate).
```
alanshaw, line 183:
```suggestion
Broker MUST issue a signed receipt to acknowledge the received request. Issued receipt MUST contain an [effect](https://github.com/ucan-wg/invocation/#7-effect) with a subsequent task (`.fx.join` field) that is run when submitted aggregate is processed and either succeeds (implying that aggregate was accepted and deals will be arranged) or fail (with `error` describing a problem with the aggregate).
```

alanshaw (changes requested):

w3-aggregation.md
```diff
@@ -179,7 +179,7 @@ Invoking `aggregate/offer` capability submits an aggregate to a broker service f
 ]
 ```
 
-Each entry of the decoded offers block, has all the necessary information for a Storage Provider to fetch and store a CAR file. The `link` field has the CAR CID, while the `commitmentProof` field has the required `proof` bytes by Storage Providers (for example, `commP`). The `src` field of each piece MUST be set to a (alphabetically sorted) list of URLs from which it can be fetched. The `size` field MUST be set to the byte size of the CAR file.
+Each entry of the decoded offers block, has all the necessary information for a Storage Provider to fetch and store a CAR file. The `link` field has the CAR CID, while the `commitmentProof` field has the required `proof` bytes by Storage Providers (for example, `commP`). The `size` field MUST be set to the byte size of the CAR file. The `src` field of each piece MUST be set to a (alphabetically sorted) list of URLs from which it can be fetched. Note that `src` field is optional and can be provided in a different part of the flow such as when deal is signed or through a previously agreed API.
```
ribasushi, line 182: Hey folks, this is concerning. The `byte`/`payload` size of the CAR file is not something ♠️ cares about nor can validate in any shape or form. Let's remove this and other `payload size` mentions from the spec 😅

vasco-santos: @ribasushi good callout. It is true that Spade does not care about this and won't validate. 

We can validate this in our side if we think that is important (perform HEAD requests to the src URLs). Main goal of having the size in an invocation is that it allows us to track metrics for data that we are requesting to Spade over time. All our metrics are being built around the information within invocations. 

With that in mind, is there any strong opinion against keeping size?

ribasushi: You can keep it on your end, but there is very strong desire on the side of ♠️ to never receive/parse/store this information, and thus never be able to return it back to you on request.

So having something as a `MUST` when on my end I have `MUST DISCARD` is... odd ;)

vasco-santos: yeah, spade side won't need to do anything at it, even at the `spade-proxy` side. Just our UCAN Stream log processor will inspect these UCAN invocations and compute metrics over time from them.

w3-aggregation.md
```diff
@@ -0,0 +1,397 @@
...
+# Terminology
+
+## Roles
+
+There are several roles in the authorization flow:
+
+| Name        | Description |
+| ----------- | ----------- |
+| Storefront | [Principal] identified by [`did:web`] identifier, representing a storage aggregator like w3up |
+| Authority   | [Principal] that represents service provider that executes invoked capabilities |
```
alanshaw, line 34:
```suggestion
| Authority   | [Principal] that represents the service provider that executes invoked capabilities |
```

alanshaw: Is this the "Broker" mentioned below?

w3-aggregation.md
```diff
@@ -0,0 +1,397 @@
...
+
+# Protocol
+
+## Overview
+
+A Storefront is the entry point for user/application data into web3. It will act on behalf of users to move data around into different storage points. One of the key storage presences may be Filecoin Storage Providers.
+
+### Authorization
+
+Broker MUST have an authorization mechanism for allowed Storefront principals (e.g. web3.storage). Either by out-of-band exchange of information or through a well defined API. In other words, a broker can authorize invocations from `did:web:web3.storage` by validating the signature is from the DID. This way, it allows web3.storage to rotate keys and/or re-delegate access without having to coordinate with the broker.
```
alanshaw, line 59:
```suggestion
Broker MUST have an authorization mechanism for allowed Storefront principals (e.g. web3.storage). Either by out-of-band exchange of information or through a well defined API. For example, a broker can authorize invocations from `did:web:web3.storage` by validating the signature is from the DID. This way, it allows web3.storage to rotate keys and/or re-delegate access without having to coordinate with the broker.
```

alanshaw (approved):

w3-aggregation.md
vasco-santos: yes

PR #57, opened 2023-03-07
feat: add UCAN Filecoin aggregations spec

alanshaw:
WIP

### DIFF
--- a/w3-aggregations.md
+++ b/w3-aggregations.md
@@ -1,0 +1,76 @@
+# UCAN Filecoin aggregations
+
+This document describes a simple UCAN service allowing an implementor to receive an aggregate of CAR files for inclusion in a Filecoin deal.
+
+## Capabilities
+
+The service exposes the following invokeable capabilities:
+
+### `aggregate/add`
+
+```json
+{
+  "iss": "did:web:web3.storage",
+  "aud": "did:web:spade.storage",
+  "att": [{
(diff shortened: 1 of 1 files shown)


olizilla:
I like it. It would be rad if the aggregate/add with: resource was a DID for a regular w3 space... and we create a new space for it that we use for the general case of system-wide, car-from-anyone, aggregates...

...but we allow for the possibility of using other user-created spaces for it, where the tooling lets users create custom aggregates that only have their own CARs in, and the metadata is stored in a space they control.

w3-aggregations.md
```diff
@@ -0,0 +1,76 @@
...
+
+Informs the service that an aggregate of CARs is ready for a Filecoin deal. Successful response implies that the aggregate has been accepted by the service to be added to a Filecoin deal.
+
+The `nb.aggregate` value is a CAR CID that refers to a "Ferry" aggregate, a collection of `dag-cbor` blocks with format:
+
+```json
+{
+  "link": { "/": "bag..." },
+  "size": 110101,
+  "md5": "md5...",
```
vasco-santos, line 33: we need to sync with riba on the need of this. It can potentially be used with client to guarantee data integrity. However, I assume spade will validate `commP`. So, data integrity will be checked there anyway.

Gozala:
Closing this in favor of #51

PR #60, opened 2023-04-17
feat: add merkle clock spec

alanshaw:
This specification describes a method of managing a [Merkle Clock](https://arxiv.org/pdf/2004.00107.pdf) of partially ordered events via UCAN invocations.

### DIFF
--- a/w3-clock.md
+++ b/w3-clock.md
@@ -1,0 +1,82 @@
+# Merkle Clock
+
+![wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+## Editors
+
+* [Alan Shaw](https://github.com/alanshaw), [DAG House](https://dag.house/)
+
+## Authors
+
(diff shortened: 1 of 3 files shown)


w3-clock.md
```diff
@@ -0,0 +1,82 @@
...
+  parents [&Event]
+  data &Data
+}
+
+type Data = Any
+```
+
+## Advancing a clock
+
+An agent may invoke a `clock/advance` capability to advance the head of a merkle clock.
```
vasco-santos, line 41: what if we call it `clock/tick`?

alanshaw: I don't have _super_ strong feelings but IMO "tick" _suggests_ a bit more that this is a traditional "global time" based clock that will tick periodically, however merkle clocks are logical clocks that encode causal information - time only advances for them when new events are added, hence "advance" being the verb here.

alanshaw:
Merging as this spec describes what is currently deployed: https://github.com/web3-storage/w3clock

PR #63, opened 2023-06-10
feat: CARv2 MultiIndex

alanshaw:
A multi-index is a CARv2 index format that allows multiple CARv2 indexes to be grouped together in a single index.

### DIFF
--- a/CARv2 MultiIndex.md
+++ b/CARv2 MultiIndex.md
@@ -1,0 +1,40 @@
+# CARv2 MultiIndex
+
+![wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+## Editors
+
+* [Alan Shaw](https://github.com/alanshaw), [DAG House](https://dag.house/)
+
+## Authors
+
+* [Alan Shaw](https://github.com/alanshaw), [DAG House](https://dag.house/)
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Abstract
+
+A multi-index is a CARv2 index format that allows multiple CARv2 indexes to be grouped together in a single index.
+
+Typically an index with this format is _not_ stored _in_ a CAR file but as a side index.
+
+## Format `0x0402`: MultiIndex
+
+An unsigned varint of `0x0402` at the "Index offset" byte position or more typically for this type of index, at the start of the file, indicates the following bytes should be interpreted as the "MultiIndex" format.
+
+The format allows multiple CARv2 indexes to be grouped together in a single index, and takes the following form:
+
+```
+| 0x0402 | count (uint32) | car-multihash | carv2-index | car-multihash | carv2-index | ... |
+```
+
+Immediately following the codec:
+
+1. `count` - the number of CAR indexes contained in this multi index.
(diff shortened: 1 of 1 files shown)


CARv2 MultiIndex.md
```diff
@@ -0,0 +1,40 @@
...
+Typically an index with this format is _not_ stored _in_ a CAR file but as a side index.
+
+## Format `0x0402`: MultiIndex
+
+An unsigned varint of `0x0402` at the "Index offset" byte position or more typically for this type of index, at the start of the file, indicates the following bytes should be interpreted as the "MultiIndex" format.
+
+The format allows multiple CARv2 indexes to be grouped together in a single index, and takes the following form:
+
+```sh
+| 0x0402 | count (uint32) | car-multihash | carv2-index | car-multihash | carv2-index | ... |
```
olizilla, line 30: why a `uint32` here and not a `varint`? wouldn't a unit32 _have_ to be encoded as 32 bits, while a varint would be more compact for smaller numbers?

alanshaw: Yes it'll probably be a low number most of the time in our use case so lets save some bytes!

CARv2 MultiIndex.md
```diff
@@ -0,0 +1,40 @@
...
+
+* [Alan Shaw](https://github.com/alanshaw), [DAG House](https://dag.house/)
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Abstract
+
+A multi-index is a CARv2 index format that allows multiple CARv2 indexes to be grouped together in a single index.
```
travis, line 19: I may be missing something and/or this might grow on me, but there's a pedant inside me saying that usually when we talk about multi* things in the IPFS ecosystem (ie, multihash, multiaddress) we're usually talking about a string of bytes that may be one of a number of different formats, and so when I glanced at this the first thing I thought was "oh this is about a thing that lets you disambiguate different strings of bytes that represent encoded indices" rather than "this is a thing that concatenates different indices together."

Not remotely a blocker, but might be worth clarifying or trying to find a different name - at the very least I just want to confirm that this isn't a multi* thing in the sense of multihashes or multiaddresses, is that right?

travis: Bonus points if you can work a Voltron or Power Rangers reference into the name since this is sort of like a Megazord for indices? 
image

alanshaw: > Not remotely a blocker, but might be worth clarifying or trying to find a different name - at the very least I just want to confirm that this isn't a multi* thing in the sense of multihashes or multiaddresses, is that right?

Yes that is correct - I'm open to a better name suggestion 🙂

`CombinedIndex`, `AggregateIndex`...? ...not sure. Want to express that it's not just a joining, since it also carries information about what CAR each index pertains to.

vasco-santos: This is a good feedback, I also thought about the naming creating some confusion. Could `DagIndex` or would we want to use this out of the context of representing a Dag?  I think that naming could work well given it can be an index for either a root CID having all its blocks, as well as sub-dags that could be optimized.

alanshaw: @travis  I could get behind `ZordIndex`...

alanshaw: * `CombinedIndex` -  join or merge to form a single unit or substance
* `CompositeIndex` - made up of several parts or elements
* `CompoundIndex` - a thing that is composed of two or more separate elements

Would be good to get a name sorted - I'd like to merge this as we're using it in production for rollup indexes in freeway.

PR #66, opened 2023-07-17
fix: aggregation spec to use height instead of size

vasco-santos:
Renames `size` to `height` for `piece`/`aggregate` based on described in https://github.com/web3-storage/data-segment/pull/13

Note that:
- given `piece` and `aggregate` are both perfect binary trees, leaf count of can be derived as `2 ** height`, and the size of the tree can be derived by `leafCount * Node.Size` (32).

### DIFF
--- a/w3-aggregation.md
+++ b/w3-aggregation.md
@@ -132,7 +132,7 @@ A Storefront principal can invoke a capabilty to offer an aggregate that is read
       "offer": { "/": "bafy...many-cars" }, /* dag-cbor CID with offer content */
       "piece": {
         "link": { "/": "commitment...aggregate-proof" },
-        "size": 10102020203013342343
+        "height": 4 /* height of the perfect binary tree for the aggregate */
       } /* commitment proof for aggregate */
     }
   }],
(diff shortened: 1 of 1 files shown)


w3-aggregation.md
```diff
@@ -343,10 +345,13 @@ type AggregateOfferDetail struct {
 
 type Offer [PieceInfo]
 
+# Adapted from `PieceInfo` type in filecoin
 # https://github.com/filecoin-project/go-state-types/blob/1e6cf0d47cdda75383ef036fc2725d1cf51dbde8/abi/piece.go#L47-L50
+# Uses `height` field instead of `size`. `height` field can be used to derive `leafCount` and consequently `size`, while
+# allowing the usage of smaller numbers instead of `bigint`.
 type PieceInfo {
```
vasco-santos, line 352: @Gozala by this change, I am considering to just rename `PieceInfo` to `PieceProof` instead.

Motivation would be to create a new name instead of relying on `PieceInfo` that is already known to have `cid` + `size`. We could just name it `PieceProof` and add comment saying it can be used to derive `PieceInfo`. What do you think?

w3-aggregation.md
```diff
@@ -132,7 +132,7 @@ A Storefront principal can invoke a capabilty to offer an aggregate that is read
       "offer": { "/": "bafy...many-cars" }, /* dag-cbor CID with offer content */
       "piece": {
```
vasco-santos, line 133: should this be `pieceProof` or even just `proof`? See rationale on other review comment

w3-aggregation.md
alanshaw: IMO it's fine to leave as `PieceInfo`, the point is for this information to now be the "piece information" as we'd not materialize and pass around a size value unless for display.

alanshaw (approved):
LGTM - I think a minor clarification on "perfect" binary tree would help but I _think_ the meaning is obvious so not a blocker.

w3-aggregation.md
```diff
@@ -141,14 +141,16 @@ A Storefront principal can invoke a capabilty to offer an aggregate that is read
...
+Invoking `aggregate/offer` capability submits an aggregate to a broker service for inclusion in one or more Filecoin deals. The `nb.piece` field represents the proof of the `piece` to be offered for the deal. It contains the proof CID `piece.link` together with the `piece.height` of the perfect binary tree computed for this aggregate. The `height` can be used to derive the leaf count (`2 ** height`), which can then be used to derive the `size` of the piece (`leafCount * Node.Size` where `Node.Size` is 32).
+
+The `nb.offer` field represents a "Ferry" aggregate offer that is ready for a Filecoin deal. Its value is the DAG-CBOR CID that refers to a "Ferry" offer. It encodes a dag-cbor block with an array of entries representing all the CAR files to include in the aggregated deal. This block MUST be included in the CAR file that transports the invocation. Its format is:
```
Gozala, line 146: Do we expect CARs in `offer` to be in same order as they appear in the aggregate ? I think we need a note whether they SHOULD, MUST, MAY or are RECOMMENDED to follow the order.

w3-aggregation.md
Gozala: I really don't want to call it a `PieceProof` as it can be confused with `InclusionProof` (in fact that's how it a appears in the go implementation).

If you want to rename perhaps call it a `PieceDetail`, personally I don't have strong feelings about the name. In fact it is a mistake that height is not part of the CID, but I'm not sure it's worth trying to fix that.

Gozala (approved):

w3-aggregation.md
vasco-santos: > Actually can we just call this an aggregate instead ?

we previously went back from `aggregate` to `piece` given they by the end of the day are the same. Have `height` and `cid`, and perhaps in the future they could be encoded in the CID as you say.

w3-aggregation.md
vasco-santos: yes, they MUST because other party might intend to validate before submitting . I will add

PR #67, opened 2023-07-18
feat: w3-deal protocol

Gozala:
First pass on the https://github.com/web3-storage/w3filecoin/issues/29 flow. 

- Translates structs used by https://github.com/data-preservation-programs/filsigner-relayed/ to IPLD Schema / TS interfaces
- Describes HTTP API flow that could be implemented to allow spade to invoke signing without learning about UCANs
   - We can also route through spade-proxy as an alternative or complimentary approach

### DIFF
--- a/w3-deal.md
+++ b/w3-deal.md
@@ -1,0 +1,217 @@
+# Filecoin Deal Signing Protocol
+
+![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+## Editors
+
+- [Irakli Gozalishvili], [Protocol Labs]
+
+## Authors
+
+- [Irakli Gozalishvili], [Protocol Labs]
+
+# Abstract
+
+This specification describes a [UCAN] protocol allowing a broker like [spade] to request signing a storage deal.
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+# Terminology
+
+## Roles
+
+There are several roles that actors can have in this protocol:
+
+| Name        | Description |
+| ----------- | ----------- |
+| Storefront | [Principal] identified by [`did:web`] identifier, representing a storage aggregator like w3up |
+| Broker   | [Principal] that arranges filecoin deals with storage providers like [spade] |
+| Agency | [Principal] submitting pieces to the _Broker_ like spade-proxy |
+
+### Storefront
+
+A _Storefront_ is a type of [principal] identified by a [`did:web`] that aggregates user data into [aggregate][] pieces and submits those to the broker to arrange deals with storage providers.
+
+### Broker
+
+A _Broker_ is a type of [principal] identified that arranges deals for the aggregates submitted by _Storefront_.
+
+# Protocol
+
+## Overview
+
+All the filecoin deals need to be signed by a Fil wallet, in order to avoid passing private keys to wallet _Storefront_ could delegate a capability to a sign a deal to a _Broker_ instead.
+
+Here we propose set of UCAN capabilities that can be invoked by authorized actors (like _Broker_) to sign deals such deals.
+
+## IPLD Schema
+
+```ipldsch
+type Deal union {
+  | Sign "deal/sign"
+} representation inline {
+  discriminantKey "can"
+}
+
+type Sign struct {
+  with    StorefrontDID
+  nb      DealProposal
+}
+
+# @see https://github.com/filecoin-project/go-state-types/blob/ff2ed169ff566458f2acd8b135d62e8ca27e7d0c/builtin/v9/market/deal.go#L201-L221
+# We use capital case letters because that the way go likes them
+type DealProposal struct {
+  PieceCID              &Piece
+  PieceSize             PieceSize
+  VerifiedDeal          bool
+  # Signer wallet (f0) address
+  Client                Address
+  # Storage provider wallet (f0) address for whom the contract is made
+  Provider              Address
+  # Label is an arbitrary client chosen label to apply to the deal
+  Label                 DealLabel
+
+  # Nominal start epoch. Deal payment is linear between StartEpoch and EndEpoch,
+  # with total amount StoragePricePerEpoch * (EndEpoch - StartEpoch).
+  # Storage deal must appear in a sealed (proven) sector no later than StartEpoch,
+  # otherwise it is invalid.
+  StartEpoch            ChainEpoch
+  EndEpoch              ChainEpoch
+  StoragePricePerEpoch  TokenAmount
+
+  ProviderCollateral    TokenAmount
+  ClientCollateral      TokenAmount
+}
+
+type StorefrontDID string
+
+# Piece CID is Piece / Aggregate merkle root encoded as CID 
+type Piece = Any
+
+# @see https://github.com/filecoin-project/go-state-types/blob/ff2ed169ff566458f2acd8b135d62e8ca27e7d0c/abi/piece.go#L12
+type PaddedPieceSize = Uint64
+
+# @see https://github.com/filecoin-project/go-address/blob/master/address.go#L39-L40
+type Address struct { addr: string }
+
+
+# The DealLabel is a kinded union of string or byte slice.
+# It serializes to a CBOR string or CBOR byte string depending on which form it
+# takes.
+# The zero value is serialized as an empty CBOR string (maj type 3).
+# @see https://github.com/filecoin-project/go-state-types/blob/ff2ed169ff566458f2acd8b135d62e8ca27e7d0c/builtin/v9/market/deal.go#L37C1-L44C1
+type DealLabel struct {
+ bs        Bytes
+ notString bool
+}
+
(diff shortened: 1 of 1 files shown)


w3-deal.md
```diff
@@ -0,0 +1,217 @@
...
+
+### `deal/sign`
+
+Broker can invoke `deal/sign` capability with `DealProposal` in (`nb` field). Storefront MUST encode supplied `DealProposal` (`nb` field) into a CBOR block and then sign it with a wallet private key.
+
+Provider MUST response with raw bytes of the signature.
+
+## HTTP Interface
+
+Given that some actors (e.g. Spade) do not support UCANs natively they are not able to send signed invocations. As compromise protocol implementer is RECOMMENDED to expose plain HTTP API that tradeoffs some security for convenience of interop.
```
Gozala, line 186:
```suggestion
Given that some actors (e.g. Spade) do not support UCANs natively they are not able to send signed invocations. As compromise protocol implementer is RECOMMENDED to expose plain HTTP API that trade-offs some security for convenience of interop.
```

w3-deal.md
```diff
@@ -0,0 +1,217 @@
...
+
+> ℹ️ Since ♠️ Spade does not support UCANs (yet), _Agency_ could instead create an invocation UCAN and pass it on to Spade so it could be used by spade as a plain, but short lived, JWT token for signing that specific deal.
+
+### Long term authorization
+
+_Storefront_ (w3up) MAY delegate unconstrained `deal/sign` UCAN capability to the _Agency_ (spade-proxy). By leaving out `Piece` and `Size` fields it will authorize it to sign any deals.
+
+_Agency_ could also re-delegate same unconstrained `deal/sign` UCAN capability to the _Agency_ (spade-proxy) allowing it to sign any deals.
+
+This tradeoffs increased security for convenience.
```
Gozala, line 208:
```suggestion
This trade-offs increased security for convenience.
```

w3-deal.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Broker
+
+A _Broker_ is a type of [principal] identified that arranges deals for the aggregates submitted by _Storefront_.
+
+# Protocol
+
+## Overview
+
+All the filecoin deals need to be signed by a Fil wallet, in order to avoid passing private keys to wallet _Storefront_ could delegate a capability to a sign a deal to a _Broker_ instead.
```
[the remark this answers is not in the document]
vasco-santos: ok, I see now the HTTP interface section below. With that in mind, I would suggest to add a small note here still to avoid same confusion from readers until later

w3-deal.md
```diff
@@ -0,0 +1,217 @@
...
+
+Provider MUST response with raw bytes of the signature.
+
+## HTTP Interface
+
+Given that some actors (e.g. Spade) do not support UCANs natively they are not able to send signed invocations. As compromise protocol implementer is RECOMMENDED to expose plain HTTP API that trade-offs some security for convenience of interop.
+
+Implementation MUST expose HTTP POST endpoint that accepts `application/vnd.ucan.cbor` requests with CBOR encoded as payload. These requests MUST provide `Authorization: Bearer` header with a UCAN delegation authorizing a request. Receiving principal MUST derive invocation from the provided `Authorization` and set invoked capability `nb` field to decoded CBOR block of the payload. Receiving principal MUST execute received capability and encode result of the invocation as an HTTP response.
+
+> Above HTTP interface could be utilize by spade to obtain signatures from w3up without having to proxy them through UCAN proxy.
```
vasco-santos, line 190:
```suggestion
> Above HTTP interface could be utilized by spade to obtain signatures from w3up without having to proxy them through UCAN proxy.
```

w3-deal.md
```diff
@@ -0,0 +1,217 @@
...
+
+> ℹ️ Since ♠️ Spade does not support UCANs (yet), _Agency_ could instead create an invocation UCAN and pass it on to Spade so it could be used by spade as a plain, but short lived, JWT token for signing that specific deal.
+
+### Long term authorization
+
+_Storefront_ (w3up) MAY delegate unconstrained `deal/sign` UCAN capability to the _Agency_ (spade-proxy). By leaving out `Piece` and `Size` fields it will authorize it to sign any deals.
+
+_Agency_ could also re-delegate same unconstrained `deal/sign` UCAN capability to the _Agency_ (spade-proxy) allowing it to sign any deals.
+
+This trade-offs increased security for convenience.
```
vasco-santos, line 208: We can delegate capability when we perform `aggregate/offer` individually per piece (aggregate). When `spade-proxy` receives the offer it can re-delegate to the broker. Given we can easily have this, should we consider to have it required instead of the long term solution?

w3-deal.md
```diff
@@ -0,0 +1,217 @@
...
+
+> Above HTTP interface could be utilize by spade to obtain signatures from w3up without having to proxy them through UCAN proxy.
+
+## Interaction Flow
+
+### Authorization per aggregate
+
+_Storefront_ (w3up) MAY delegate `deal/sign` UCAN capability to the _Agency_ (spade-proxy) and specify `Piece` and `Size` fields of the submitted aggregate.
+
+_Agency_ (spade-proxy) could also re-delegate that capability to the _Agency_ (spade) allowing it to request signature directly from _Storefront_ front.
```
vasco-santos, line 198:
```suggestion
_Agency_ (spade-proxy) could also re-delegate that capability to the _Broker_ (spade) allowing it to request signature directly from _Storefront_ (w3up).
```

w3-deal.md
```diff
@@ -0,0 +1,217 @@
...
+  | SigningError // WalletSignError
+```
+
+## Capabilities
+
+### `deal/sign`
+
+Broker can invoke `deal/sign` capability with `DealProposal` in (`nb` field). Storefront MUST encode supplied `DealProposal` (`nb` field) into a CBOR block and then sign it with a wallet private key.
+
+Provider MUST response with raw bytes of the signature.
```
vasco-santos, line 182:
```suggestion
Provider MUST respond with raw bytes of the signature.
```

w3-deal.md
```diff
@@ -0,0 +1,217 @@
...
+type Sign struct {
+  with    StorefrontDID
+  nb      DealProposal
+}
+
+# @see https://github.com/filecoin-project/go-state-types/blob/ff2ed169ff566458f2acd8b135d62e8ca27e7d0c/builtin/v9/market/deal.go#L201-L221
+# We use capital case letters because that the way go likes them
+type DealProposal struct {
+  PieceCID              &Piece
+  PieceSize             PieceSize
```
vasco-santos, line 67:
```suggestion
  PieceSize             PaddedPieceSize
```

vasco-santos (commented):
This iteration looks great! Thanks for putting it together

1. we should present it to spade team and gather feedback sooner rather than later, so that we can understand if this is something that we can work together to land (including spade side of things). if not the case, we should reduce scope for now to only have interaction between `storefront` and `agency` in a perspective that someday they will be the same.
2. should we spec how we would provide source URLs for each piece in this flow? could be good to have something in place we could also present their team to initiate dialog on that side of things too

w3-deal.md
Gozala, line 45:
```suggestion
All the filecoin deals need to be signed by a Fil wallet, in order to avoid sharing wallet private keys, _Storefront_ could delegate a capability to a sign a deal to an _Agency_ or a _Broker_ arranging a deal on their _Storefront_s behalf.
```

How does this sound ?

w3-deal.md
```diff
@@ -0,0 +1,217 @@
...
+
+A _Broker_ is a type of [principal] identified that arranges deals for the aggregates submitted by _Storefront_.
+
+# Protocol
+
+## Overview
+
+All the filecoin deals need to be signed by a Fil wallet, in order to avoid passing private keys to wallet _Storefront_ could delegate a capability to a sign a deal to a _Broker_ instead.
+
+Here we propose set of UCAN capabilities that can be invoked by authorized actors (like _Broker_) to sign deals such deals.
```
Gozala, line 47:
```suggestion
Here we propose set of UCAN capabilities that can be invoked by authorized actors (like _Agency_ or a _Broker_) to sign deals on behalf of the delegate (_Storefront_).
```

w3-deal.md
```diff
@@ -0,0 +1,217 @@
...
+
+_Agency_ (spade-proxy) could also re-delegate that capability to the _Agency_ (spade) allowing it to request signature directly from _Storefront_ front.
+
+> ℹ️ Since ♠️ Spade does not support UCANs (yet), _Agency_ could instead create an invocation UCAN and pass it on to Spade so it could be used by spade as a plain, but short lived, JWT token for signing that specific deal.
+
+### Long term authorization
+
+_Storefront_ (w3up) MAY delegate unconstrained `deal/sign` UCAN capability to the _Agency_ (spade-proxy). By leaving out `Piece` and `Size` fields it will authorize it to sign any deals.
+
+_Agency_ could also re-delegate same unconstrained `deal/sign` UCAN capability to the _Agency_ (spade-proxy) allowing it to sign any deals.
```
Gozala, line 206:
```suggestion
_Agency_ (spade-proxy) could also re-delegate same unconstrained `deal/sign` UCAN capability to the _Broker_ (spade) allowing it to sign any deals.
```

w3-deal.md
Gozala: I think we should go for auth per offer, that said I think it's good to have both options listed. There is another tradeoff that I failed to communicate here, which is token per offer means that those tokens are likely to be included in requests and consequently can get leaked very easily. Long term tokens are likely exchanged ahead of time and out of bound, so they are less likely to get leaked in comparison to per-request approach.

Issue #69, opened 2023-07-24
aggregate offer nb fields


vasco-santos:
An aggregate can include a field for a label that can be used to encode some metadata that is meaningful to us. Note up to a limit https://github.com/filecoin-project/go-state-types/blob/master/builtin/v11/market/policy.go#L29-L30

```go
// DealMaxLabelSize is the maximum size of a deal label.
const DealMaxLabelSize = 256
```

TBD: do we need to send a `client` wallet address, endpoint, ?

PR #70, opened 2023-07-24
fix: aggregate spec offer complete

vasco-santos:
A few items are still needed to have an offer to be part of a contract with a SP. Namely, client MUST provide a deal wallet signer address to be used to sign/pay for the deal. 

An arbitrary label can also be used with metadata from the deal.

Closes https://github.com/web3-storage/specs/issues/69


vasco-santos:
@Gozala and @alanshaw asked review for this PR by mistake instead of #71. This we are still on discussions and as draft

vasco-santos:
Included in https://github.com/web3-storage/specs/pull/71

PR #74, opened 2023-08-10
feat: add spec for rate limits

travis:
ported from https://hackmd.io/@tvpl/H1whErCI3

### DIFF
--- a/w3-rate-limit.md
+++ b/w3-rate-limit.md
@@ -43,9 +43,9 @@ In other words, if an agent has a delegation for `rate-limit/*` for a given prov
 
 #### rate-limit/add
 
-Given a subject ID (eg, a`did:mailto`, a URL, a domain name, etc), set a rate limit for the entity represented by that ID. The semantics of both the subject and rate are intentionally abstract, and the service is expected to record them without much processing.
+Given a subject ID (e.g., a`did:mailto`, a URL, a domain name, etc), set a rate limit for the entity represented by that ID. The semantics of both the subject and rate are intentionally abstract, and the service is expected to record them without much processing.
 
-Consumers of rate limits are expected to query the underlying datastore where they are stored for the subjects they care about and interpret the `rate` value in a way that makes sense for their usecase.
+Consumers of rate limits are expected to query the underlying data store where they are stored for the subjects they care about and interpret the `rate` value in a way that makes sense for their use case.
 
 Returns an ID that can be used to remove this limit later.
 
@@ -87,7 +87,7 @@ export const add = capability({
 
 #### rate-limit/list
 
-Given a subject ID (eg, a`did:mailto`, a URL, a domain name, etc), list all rate limits that apply to the given subject.
+Given a subject ID (e.g., a`did:mailto`, a URL, a domain name, etc), list all rate limits that apply to the given subject.
 
 ##### inputs
 
--- a/w3-rate-limit.md
+++ b/w3-rate-limit.md
@@ -18,9 +18,9 @@ from using their service. We describe a set of capabilities for tracking and adm
 - [Capabilities](#capabilities)
   - [`rate-limit/` namespace](#rate-limit-namespace)
     - [`rate-limit/*`](#rate-limit)
-    - [`rate-limit/add`](#ratelimitadd)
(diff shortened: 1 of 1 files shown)


w3-rate-limit.md
```diff
@@ -0,0 +1,169 @@
...
+
+![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+## Editors
+
+- [Travis Vachon], [Protocol Labs]
+
+## Authors
+
+- [Travis Vachon], [Protocol Labs]
```
olizilla, line 11: needs link targets

olizilla (changes requested):
Do we have an example use-case where we would apply multiple rate-limits to a single subject?

Right now we have a multiple global rate limits on dag.house requests, which differ by request count and time interval. I think we'd need to add another property to the rate limit object defined here to be able to recreate that.

That's probably out of scope for now, but seems like a useful data point for iterating on this spec.

In the meantime we should define how multiple rate limits on the same subject should be applied, or explicitly say that it's undefined and go nuts.

travis:
Hm good question, and maybe one for @dchoi27 as he has more perspective on the types of rate limits we're likely to add in the future...

@dchoi I'd been imagining situations where, say, a person doing fraud detection added a rate limit to a space while they reviewed some problematic uploads and a separate person responsible for billing adding a rate limit to a space when it the space used more than its quota - in that case I imagine we might even want two identical rate limits (say, full blocks).

I haven't considered a case where we might want multiple rate limits with different units - request count and time interval as you say @olizilla - and I'm not sure that's even possible with the current version of this spec - @Gozala I'm not sure we designed for this, any thoughts on how we might tweak this to allow for rate limits with defined units? Maybe just a `type` field?

w3-rate-limit.md
Gozala:
```suggestion
- [Travis Vachon](https://github.com/travis), [Protocol Labs](https://protocol.ai/)

## Authors

- [Travis Vachon](https://github.com/travis), [Protocol Labs](https://protocol.ai/)
```

Gozala:
I think @olizilla is brought up interesting questions, as in you may have multiple policies applied to the same subject. I have not really thought of it, mostly I had been trying to steer us from boolean thinking (blocked/unblocked) into wider spectrum of possibilities where "blocked" is `rate: 0` and unblocked is `rate: Number.MAX_SAFE_INTEGER`. I though leaving out frame within which rate is enforced unspecified would give us flexibility to do so in the future.

I am realizing now however that we may have multiple different "frame"'s at play at the same time. In fact this reminds me of client [backoff code I had to write in the past](https://github.com/nftstorage/nft.storage/blob/af7524fbf1240b804bbbe356afb90ec4d2bd86b1/packages/niftysave/src/retry.js) because [p-retry](https://www.npmjs.com/package/p-retry) had no way to account for multiple constraints. Makes me wonder if we'll also need an exponential backoff to rebalance throughput.

Musing aside I think there are couple of options we could consider here:

1. Add another field so we have triple like`(subject, policy, rate)`. _Maybe this is a good idea triple stores RDF and Datalog/Datomic are able to model pretty much anything through such triples, we could probably even rename to (subject, predicate, value) :P_
2. We could capture that extra detail in the subject, e.g. in UCANs we use URLs for resources just so that schema / protocol can signal context. Although if we have to parse it every time it's better to just have a dedicated field.
3. Call it a day and introduce new field in the future. Downside is we'll have some records without such fields that will have to then interpret to mean some policy implicitly.

I have slight preference towards no 1, just because of triple store precedence.

Gozala:
> Right now we have a multiple global rate limits on dag.house requests, which differ by request count and time interval. I think we'd need to add another property to the rate limit object defined here to be able to recreate that.

@olizilla am I understanding it correctly that example of what we're missing here would be:

Say we have following policies in place

1. Max 100 uploads per minute
2. Max 1000 uploads per hour

If you user had been doing 100 request each minute after 10 mins we would deny requests for 50 mins. Notably if we had just second policy in place user could potentially do 200 requests per second which combination of both limits prevents.

Gozala (approved):
I suggest we merge this and address points raised by @olizilla in a followup

PR #77, opened 2023-09-08
feat: propose `admin/upload/inspect` and `admin/store/inspect` capabilities

travis:
At a high level, these are designed to enable administrators like @dchoi27 to figure out which space a given CID is associated with.

Problem CIDs come to us from a variety of sources - sometimes they are content CIDs that correspond to a particular `upload/add` invocation and sometimes they are "shard" CIDs that identify one of the CARs stored using `store/add` that make up the shards of a larger upload. `admin/upload/inspect` returns information about the upload and `admin/store/inspect` returns information about particular shards, so given an unknown CID administrators should check both.

### DIFF
--- a/w3-admin.md
+++ b/w3-admin.md
@@ -20,8 +20,8 @@ to administrative users by creating delegations signed with the service signer's
   - [`consumer/get`](#consumerget)
   - [`customer/get`](#customerget)
   - [`subscription/get`](#subscriptionget)
-  - [`trace/upload/get`](#traceuploadget)
-  - [`trace/store/get`](#tracestoreget)
+  - [`trace/upload/add`](#traceuploadadd)
+  - [`trace/store/add`](#tracestoreadd)
 
 ## Language
 
@@ -151,7 +151,7 @@ export const get = capability({
 })
 ```
 
-### `trace/upload/get`
+### `trace/upload/add`
 
 Get information about a content CID. This does not return the actual data identified by the CID, just metadata our
 system tracks, e.g. the spaces the content identified by a given CID has been uploaded to and the dates the uploads happened.
@@ -171,7 +171,7 @@ with the date it was uploaded.
 }
 ```
 
-### `trace/store/get`
+### `trace/store/add`
 
 Get information about a shard (i.e., a CAR that contains part of an upload) CID. This
 does not return the actual data identified by the CID, just metadata our system tracks,
--- a/w3-admin.md
+++ b/w3-admin.md
@@ -20,7 +20,8 @@ to administrative users by creating delegations signed with the service signer's
   - [`consumer/get`](#consumerget)
   - [`customer/get`](#customerget)
   - [`subscription/get`](#subscriptionget)
-  - [`cid/get`](#cidget)
+  - [`root/get`](#rootget)
+  - [`shard/get`](#shardget)
 
 ## Language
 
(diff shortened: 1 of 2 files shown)


w3-admin.md
```diff
@@ -147,3 +148,25 @@ export const get = capability({
...
+
+#### returns
+
+The return type currently consists of two lists: "uploads" and "stores". These lists correspond to
+the `upload/*` and `store/*` capabilities described in the [storage and upload protocol](./w3-store.md).
+
+```typescript
+{
+  uploads: {space: SpaceDID, insertedAt: Date}[]
+  stores: {space: SpaceDID, insertedAt: Date}[]
```
alanshaw, line 170:
```suggestion
  uploads: Array<{ space: SpaceDID, insertedAt: Date }>
  stores: Array<{ space: SpaceDID, insertedAt: Date }>
```

w3-admin.md
```diff
@@ -147,3 +148,25 @@ export const get = capability({
   },
 })
 ```
+
+### `cid/get`
+
+Get information about a CID. This does not return the actual data identified by the CID, just metadata our
+system tracks, e.g. the spaces the content identified by a given CID has been uploaded to and the
```
alanshaw, line 155: Hmm, to check my understanding, this is a list of spaces a shard has been stored in and/or registered as an upload in?

Maybe this is a bit too generic as `cid/get`, I'd be tempted to call it something like `shard/stat`, `shard/info` or `shard/get`.

Or are we talking about a `root` CID? How does that work with `store/add` (which only deals with shard CIDs)?

Either way this is confusing and I think needs some clarification.

w3-admin.md
```diff
@@ -147,3 +148,25 @@ export const get = capability({
...
+dates the uploads happened.
+
+#### inputs
+
+`cid: CID`
```
alanshaw, line 160:
```suggestion
`shard: CID`
```

or 

```suggestion
`root: CID`
```

???

alanshaw (changes requested):

alanshaw:
Oh I should really read the PR description before reviewing the code 🤦‍♂️

I'd vote for 2 separate capabilities for querying `upload/add`s and `store/add`s!

travis:
ok cool - I will rework this with two separate capabilities!

Gozala (changes requested):
I think current naming is too generic and we probably should avoid taking up so many namespaces. Furthermore I think it would be a good idea to group these capabilities under same namespace so that `prefix/*` delegation used.

I offered some name suggestions, but not insisting on them.

alanshaw:
+1 I'm cool with `trace/store/add` and `trace/upload/add` as the capability names.

w3-admin.md
```diff
@@ -148,3 +150,44 @@ export const get = capability({
   },
 })
 ```
+
+### `trace/upload/get`
```
Gozala: I have suggested `trace/upload/add` so it corresponds to `upload/add`, meaning it's a capability to "trace upload/add" operation. Although I realize now it goes against established add, get, list, remove pattern used in other capabilities.

However `trace/upload/get` also feels counter intuitive to me even though I get the logic. On one hand it would make sense to have `upload/trace` and `store/trace`,  but that goes against prefix defining the ownership.

How about `console/upload/inspect` and `console/store/inspect` instead ? Alternatively it could be `admin/upload/inspect` or `system/upload/inspect`. I suppose `trace/upload/inspect` works too.

travis: chatted offline and agreed to go with `admin/{upload|store}/inspect`

Gozala (approved):

PR #80, opened 2023-09-13
wip: propose `admin/block/inspect` capability

travis:
@dchoi27 often needs to get information about CIDs that identify blocks packed into CARs stored. The current process for this is to calculate the multihash of the CID (eg,
`ipfs cid format -f "%m" -b base58btc bafybeicfs2nuef675li6pxrdzar4bkmdp5ktud4shubtgkenk5gwtjqde4`) and then to use that multihash to query the `blocks-cars-position` table, which contains `carpath`, `length` and `offset`. He then uses the `carpath` to determine the parent CAR's CID, which can be used to figure out which space the content has been uploaded to, allowing him to block the space and/or user who registered the space.

I'm very open to reworking this or tackling it a different way - this is meant to be an initial proposal to start a discussion on the best way to support this admin workflow.

Also tweak the return values of the `admin/store/inspect` and `admin/upload/inspect` to reflect the (imho better) names used in the implementations.


olizilla:
I'm in favour of making our life easier. My only concern is I'd like us to figure out what the guiding principle should be for deciding if a thing is an implementation detail or a specified feature of the system.

Also note that an api for `(CID) => metadata` could be part of the content-claims universe https://github.com/web3-storage/content-claims

PR #82, opened 2023-10-04
feat: add `store/get` & `upload/get` capabilities

olizilla:
These capabilities are intended for checking inclusion, or getting the additional properties we store. This will allow clients to verify all the shards for an upload have been removed before removing an upload.

## `store/get` delegation

```js
{
  can: "store/get",
  with: "did:key:abc...",
  nb: {
    link: "bag...",
  }
}
```

### Response

Error if `link` shard cid is not in space or...

```ts
interface StoreListItem {
  /** CID of the stored CAR. */
  link: string

  /** Size in bytes of the stored CAR */
  size: number

  /** Link to a related CAR, used for sharded uploads */
  origin?: string,

  /** ISO-8601 timestamp when CAR was added to the space */
  insertedAt: string,
}
```

## `upload/get` delegation

```js
{
  can: "upload/get",
  with: "did:key:abc...",
  nb: {
    root: "bafyabc..."
  }
}
```

### Response

`Error` if root is not in uploads for space or...

```ts
interface UploadListItem {
  /** Root CID of the uploaded data item. */
  root: string

  /** CIDs of CARs that contain the complete DAG for the uploaded data item. */
  shards: string[]

  /** ISO-8601 timestamp when the upload was added to the space. */
  insertedAt: string,

  /** ISO-8601 timestamp when the upload entry was last modified. */
  updatedAt: string,
}
```

---

see also: https://github.com/web3-storage/w3up/pull/942

License: MIT

### DIFF
--- a/w3-store.md
+++ b/w3-store.md
@@ -22,11 +22,13 @@ The protocol is modeled as a set of capabilities which can be fulfilled by a [pr
   - [`store/` namespace](#store-namespace)
     - [`store/*`](#store)
     - [`store/add`](#storeadd)
+    - [`store/get`](#storeget)
     - [`store/remove`](#storeremove)
     - [`store/list`](#storelist)
   - [`upload/` namespace](#upload-namespace)
     - [`upload/*`](#upload)
     - [`upload/add`](#uploadadd)
+    - [`upload/get`](#uploadget)
     - [`upload/remove`](#uploadremove)
     - [`upload/list`](#uploadlist)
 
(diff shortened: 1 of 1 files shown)


alanshaw (approved):
Looks fine but we should update all of these to not use strings for CIDs....

w3-store.md
```diff
@@ -143,6 +145,66 @@ If `status == 'upload'`, the response will include additional fields containing
...
+}
+```
+
+#### Responses
+
+*Note*: This section is non-normative and subject to change, pending the [formalization of receipts and invocations][invocation-spec-pr].
+
+Executing a `store/get` invocation with a service provider should return a response object.
+
+If a failure occurs, the response will have an `error` field with a value of `true`, and a `message` string field with details about the error.
```
travis, line 188: maybe I'm misunderstanding this, but won't the error response look like:

```js
{ error: { name: "ErrorName", message: "An error message" } }
```

?

w3-store.md
```diff
@@ -143,6 +145,66 @@ If `status == 'upload'`, the response will include additional fields containing
...
+
+On success, the response object will have the following shape:
```
travis, line 190: and similarly, I think the response will look like:

```javascript
{ ok: {link: "bagy...", size: 42, insertedAt: "2023-01-01T01:01:01" } }
```

w3-store.md
```diff
@@ -340,6 +402,68 @@ interface UploadAddResponse {
...
+
+The `with` resource URI must be set to the DID of the space to get the upload metadata from.
+
+#### Derivations
+
+`upload/get` can be derived from an [`upload/*`](#upload) or [`*`][ucan-spec-top] capability with a matching `with` resource URI.
+
+#### Caveats
+
+The `root` caveat must contain the root CID of the data item to remove.
```
travis, line 421: s/remove/return/

w3-store.md
```diff
@@ -340,6 +402,68 @@ interface UploadAddResponse {
...
+
+| `field`   | `value`                           | `required?` | `context`                                                     |
+| --------- | --------------------------------- | ----------- | ------------------------------------------------------------- |
+| `root`    | data CID string, e.g. `bafy...`   |             | The root CID of the upload to get metadata for                |
```
travis, line 425: required should be true I think

w3-store.md
```diff
@@ -340,6 +402,68 @@ interface UploadAddResponse {
...
+}
+```
+
+#### Responses
+
+*Note*: This section is non-normative and subject to change, pending the [formalization of receipts and invocations][invocation-spec-pr].
+
+Executing an `upload/get` invocation with a service provider should return a response object.
+
+If a failure occurs, the response will have an `error` field with a value of `true`, and a `message` string field with details about the error.
```
travis, line 447: ditto re: response formats in this section

travis (changes requested):
could update these post-merge of course, but I do think these changes need doing

w3-store.md
olizilla: yep,  but i think we need to make it clearer that these tables are referring to "when used as an invocation", as the `root` is not required in the schema, as folks should be allowed to make a delegation that grants you `upload/get` on any root. same for `store.get`. My gut feel is we should make the tables present the ucanto schema rather than the "when used as invocation" form.

olizilla: (yep, as i deliberately differed from the other items in this spec where they have marked an optional field as required)

w3-store.md
olizilla: ah yes you right. I was blending this in with the rest of this spec, but things have changed!

I can't find a place in our other specs where we document the expected output shapes which seems odd too.

w3-store.md
travis: ahhhh got it, I agree re: presenting ucanto schema!!

w3-store.md
travis: yeaaaa I agree, I think it's a hole in our specs that we haven't documented the input/return types across the board, maybe even a DX issue worth fixing asap?

w3-store.md
olizilla, line 421:
```suggestion
The `root` caveat must contain the root CID of the item to get metadata for.
```

olizilla:
implemented in https://github.com/web3-storage/w3up/pull/1178

PR #85, opened 2023-12-13
feat: add IPNI spec

olizilla:
Describes the `ipni/offer` capability and how to merge inclusion claims with IPNI Advertisements

> For IPNI we assert that we can provide batches of multihashes by signing "Advertisements". 
> 
> With an inclusion claim, a user asserts that a CAR contains a given set of multihashes via a car index.
> 
> This spec describes how to merge these two concepts by adding an `ipni/offer` capability to submit an inclusion claim as an IPNI Advertisement.

License: MIT

### DIFF
--- a/w3-ipni.md
+++ b/w3-ipni.md
@@ -1,0 +1,237 @@
+# W3 IPNI Protocol
+
+![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+## Authors
+
+- [olizilla], [Protocol Labs]
+
+# Abstract
+
+For IPNI we assert that we can provide batches of multihashes by signing "Advertisements". 
+
+With an inclusion claim, a user asserts that a CAR contains a given set of multihashes via a car index.
+
+This spec describes how to merge these two concepts by adding an `ipni/offer` capability to submit an inclusion claim as an IPNI Advertisement.
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Introduction 
+
+**What this unlocks** (tl;dr)
+
+- Create 1 or more IPNI Adverts per user uploaded CAR and set the ContextID to be the CAR CID (instead of arbitrary batches with no ContextId)
+    - With this we (or anyone, ipni is open access) can now use IPNI to find which CAR a block is in. The context id bytes provide the CAR CID for any block look up. The CAR CID can then be used to find the CAR index via our content-claims API.
+    - We can **delete** the IPNI records by CAR CID if the CAR is deleted.
+- Make IPNI advertising an explicit UCAN capability that clients can invoke rather than a side-effect of bucket events
+    - With this we are free to write CARs anywhere. The users agent invokes a `ipni/offer` capability to ask us to publish and IPNI ad for the blocks in their CAR.
+    - This empowers the user to opt-in or out as they need, and allows us to bill for the (small) cost of running that service.
+- Put the lime in the coconut. Put an inclusion claim in the IPNI advert metadata.
+    - We show the source of our provider claim is a user signed inclusion content claim.
+    - We have to sign IPNI Adverts as the provider, so we can warn folks that this ad is as good as the user provided content claim it includes.
+
+### Quick IPNI primer
+
+IPNI ingests and replicates billions of signed provider claims for where individual block CIDs can be retrieved from.
+
+Users can query IPNI servers for any CID, and it provides a set of provider addresses and transport info, along with a provider specific ContextID and optional metadata.
+
+http://cid.contact hosts an IPNI server that Protocol Labs maintains. *(at time of writing)*
+
+```bash
+$ curl https://cid.contact/cid/bafybeicawc3qwtlecld6lmtvsndimoz3446xyaprgsxvhd3aapwa2twnc4 -sS | jq
+```
+
+```json
+{
+  "MultihashResults": [
+    {
+      "Multihash": "EiBAsLcLTWQSx+WydZNGhjs75z18AfE0r1OPYAPsDU7NFw==",
+      "ProviderResults": [
+        {
+          "ContextID": "YmFndXFlZXJheTJ2ZWJsZGNhY2JjM3Z0em94bXBvM2NiYmFsNzV3d3R0aHRyamhuaDdvN2o2c2J0d2xmcQ==",
+          "Metadata": "gBI=",
+          "Provider": {
+            "ID": "QmQzqxhK82kAmKvARFZSkUVS6fo9sySaiogAnx5EnZ6ZmC",
+            "Addrs": [
+              "/dns4/elastic.dag.house/tcp/443/wss"
+            ]
+          }
+        },
+        {
+          "ContextID": "YmFndXFlZXJheTJ2ZWJsZGNhY2JjM3Z0em94bXBvM2NiYmFsNzV3d3R0aHRyamhuaDdvN2o2c2J0d2xmcQ==",
+          "Metadata": "oBIA",
+          "Provider": {
+            "ID": "QmUA9D3H7HeCYsirB3KmPSvZh3dNXMZas6Lwgr4fv1HTTp",
+            "Addrs": [
+              "/dns4/dag.w3s.link/tcp/443/https"
+            ]
+          }
+        }
+```
+
+web3.storage publishes the blocks it can provide by encoding a batch of multihashes as an IPLD object and writing it to S3 as an `Advertisement`, addressed by it's CID. 
+
+An `Advertisement` includes `Provider` info which claims that a the batch of multihashes are available via bitswap or HTTP, and are signed by the provider PeerId private key; Each advert is a claim that this peer will provide that batch of multihashes.
+
+Advertisements also include a CID link to any previous ones from the same provider forming a hash linked list.
+
+The latest `head` CID of the ad list can be broadcast over gossipsub, to be replicated and indexed by all listeners, or POSTed over HTTP to specific IPNI servers as a notification to pull and index the latest ads from you at their earliest convenience.
+
+The advert `ContextID` allows providers to specify a custom grouping key for multiple adverts. You can update or remove multiple adverts by specifying the same ContextID. The value is an opaque byte array as far as IPNI is concerned, and is provided in the query response.
+
+A `Metadata` field is also available for provider specific retrieval hints, that a user should send to the provider when making a request for the block, but the mechanism here is unclear (http headers? bitswap what now?). Regardless it is more space for provider specified bytes... like maybe... a content claim! *(foreshadowing!)*
+
+### How web3.storage integrates IPNI today
+
+w3s publishes IPNI advertisements as a side-effect of the e-ipfs car block indexer.
+
+Each multihash in a CAR is sent to an SQS queue. The `publisher-lambda` takes batches from the queue, encodes and signs `Advertisement`s and writes them to S3 as json.
+
+The lambda makes an http request to the cid.contact to inform it when the head CID of the Advertisement linked list changes.
+
+The cid.contact IPNI server fetches new head Advertisement from our s3 bucket, and any others in the chain it hasn't read yet, and updates it's indexes.
+
+Our `Advertisement`s contain arbitrary batches of multihashes defined by SQS queue batching config. The ContextID is set to opaque bytes (a custom hash of the hashes).
+
+#### Diagram
+
+```mermaid
+flowchart TD
+    A[(dotstorage\nbucket)] -->|ObjectCreated fa:fa-car| B(bucket-to-indexer ƛ)
+    B -->|region/bucket/cid/cid.car| C[/indexer queue/]
+    C --> indexer(Indexer ƛ)
+    indexer --> |zQmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn| E[/multihash queue/]
+    E --> F(ipni Advertisement content ƛ)
+    F --> |PUT /advertCid|I
+    F --> |advert CID| G[/Advertisement queue/]
+    G --> H(ipni publish ƛ)
+    H --> |PUT /head|I[(Advert Bucket)]
+    H --> |POST head|IPNI[["`**IPNI**`"]]
+    
+    carpark[(carpark\nbucket)] --> |ObjectCreated fa:fa-car|w3infra-carpark-consumer(carpark-consumer ƛ)
+    w3infra-carpark-consumer -->|region/bucket/cid/cid.car| C[/indexer queue/]
+
+    indexer ---> dynamo[Dynamo\nblocks index]
+```
+
+## Proposal
+
+Provide a `ipni/offer` ucan ability to sign and publish an IPNI Advertisement for the set of multihashes in a CAR a user has stored with w3s, to make them discoverable via IPFS implementations and other IPNI consumers.
+
+```mermaid
+sequenceDiagram
+    actor Alice
+    Alice->>w3s: ipni/offer (inclusion proof)
+    activate w3s
+    w3s-->>w3s: fetch & verify index
+    w3s-->>w3s: write advert
+    w3s-->>Alice: OK (advertisement CID)
+    w3s-->>ipni: publish head (CID)
+    deactivate w3s
+    ipni-->>w3s: fetch advert
+    activate ipni
+    ipni-->>ipni: index entries
+    deactivate ipni
+    Alice->>ipni: query (CID)
+```
+
+
+Invoke it with the CID for an [inclusion-claim] that associates a CAR CID wth [MultihashIndexSorted CARv2 Index] CID.
+
+:::info
+Other CAR index forms may be supported in the future. A more convenient external CAR index format would provide the offset byte and block byteLength for a multihash from the start of the CAR file.
+:::
+
+
+```json
+{
+  "iss": "did:key:zAlice",
+  "aud": "did:web:web3.storage",
+  "att": [{
+    "can": "ipni/offer",
+    "with": "did:key:space", // users space DID
+    "nb": {
+      "inclusion": CID   // inclusion claim CID
+    }
+  }]
+}
+```
+
+**Inclusion claim**
+```json
+{
+  "content": CID, // CAR CID
+  "includes": CID // CARv2 Index CID
+}
+```
+
+When `ipni/offer` is invoked the service must fetch the inclusion claim. The encoded claim block may be sent with the invocation.
+
+The service must fetch he CARv2 index and parse it to find the set of multihashes included in the CAR. see: [Verifying the CARv2 Index](#verifying-the-carv2-index)
+
+The set of multihashes must be encoded as 1 or more [IPNI Advertisements]. 
+
+```ipldsch
+type Advertisement struct {
+    PreviousID optional Link
+    Provider String
+    Addresses [String]
+    Signature Bytes
+    Entries Link
+    ContextID Bytes
+    Metadata Bytes
+    IsRm Bool
+    ExtendedProvider optional ExtendedProvider
+}
+```
+
+- `Entries` must be the CID of an `EntryChunk` for a subset (or all) of multihashes in the CAR.
+- `ContextID` must be the byte encoded form of the CAR CID.
+- `Metadata` must be the bytes of the inclusion claim.
+ 
+See: [Encoding the IPNI Advertisement](#encoding-the-ipni-advertisement)
+
+The Advertisement CID should be POSTed to an IPNI server. `cid.contact` is assumed initially.
+
+The Advertisement CID should be gossiped on the `/indexer/ingest/mainnet` topic so they can be replicated by other IPNI servers, to ensure many nodes can answer queries for the blocks we host.
+
+
+### Verifying the CARv2 Index
+
+The service must fetch the CARv2 Index and may verify 1 or more multihashes from the index exist at the specified offsets in the associated CAR. 
+
+The verifier should pick a set of multihashes at random and fetch the bytes from the CAR identified by the index entry and verify it's multihash. The invocation must return an error if any entry is found to be invalid.
+
+Random validation of a number of blocks allows us to detect invalid indexes and lets us tune how much work we are willing to do per car index.
+
+Full validation of every block is not recommended as it opens us up to performing unbounded work. *We have seen CAR files with millions of tiny blocks.*
+
+
+### Encoding the IPNI Advertisement
+
+> The set of multihashes must be encoded as 1 or more [IPNI Advertisements].
+
+Where the IPLD encoded size of an `EntryChunk` with the set of multihashes would exceed 4MiB (the upper limit for a block that can be transferred by libp2p) the set of multihashes must be split into multiple `EntryChunk` blocks
+
+```ipldsch
+type EntryChunk struct {
+    Entries [Bytes]
+    Next optional Link
+}
+```
+
+It is possible to create long chains of `EntryChunk` blocks by setting the `Next` field to the CID to another `EntryChunk`, but this requires an entire EntryChunk to be fetched and decoded, before the IPNI server can determine the next chunk to fetch.
+
+The containing CAR CID provides a useful `ContextID` for grouping multiple (light weight) Advertisement blocks so it is recommended to split the set across multiple `Advertisement` blocks each pointing to an `EntryChunk` with a partition of the set of multihashes in, and the `ContextId` set to the CAR CID.
+
+
+[MultihashIndexSorted CARv2 Index]:  https://ipld.io/specs/transport/car/carv2/#format-0x0401-multihashindexsorted
+
+[inclusion-claim]: https://github.com/web3-storage/content-claims?tab=readme-ov-file#inclusion-claim
+
+[IPNI Advertisements]: https://github.com/ipni/specs/blob/main/IPNI.md#advertisements
+
+[olizilla]: https://github.com/olizilla


w3-ipni.md
```diff
@@ -0,0 +1,237 @@
...
+    
+    carpark[(carpark\nbucket)] --> |ObjectCreated fa:fa-car|w3infra-carpark-consumer(carpark-consumer ƛ)
+    w3infra-carpark-consumer -->|region/bucket/cid/cid.car| C[/indexer queue/]
+
+    indexer ---> dynamo[Dynamo\nblocks index]
+```
+
+## Proposal
+
+Provide a `ipni/offer` ucan ability to sign and publish an IPNI Advertisement for the set of multihashes in a CAR a user has stored with w3s, to make them discoverable via IPFS implementations and other IPNI consumers.
```
vasco-santos, line 122: I wonder if we could make a goal of this to move closer to decentralize IPNS. Have a separate service that provides this capability that could be implemented by multiple parties. This would also allow us to reach out to IPNI team and see if they would like to run this service instead of us, making it available for other users of IPNS if they would like to get into the UCAN world.

we could therefore decouple w3s from this system

w3-ipni.md
```diff
@@ -0,0 +1,237 @@
...
+  "content": CID, // CAR CID
+  "includes": CID // CARv2 Index CID
+}
+```
+
+When `ipni/offer` is invoked the service must fetch the inclusion claim. The encoded claim block may be sent with the invocation.
+
+The service must fetch he CARv2 index and parse it to find the set of multihashes included in the CAR. see: [Verifying the CARv2 Index](#verifying-the-carv2-index)
+
+The set of multihashes must be encoded as 1 or more [IPNI Advertisements].
```
vasco-santos, line 175: is it really required to be 1? Could it not be 0 as well? I think the protocol should not imply that validation is required for at least one block, but that validation MAY happen for each block

w3-ipni.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Verifying the CARv2 Index
+
+The service must fetch the CARv2 Index and may verify 1 or more multihashes from the index exist at the specified offsets in the associated CAR. 
+
+The verifier should pick a set of multihashes at random and fetch the bytes from the CAR identified by the index entry and verify it's multihash. The invocation must return an error if any entry is found to be invalid.
+
+Random validation of a number of blocks allows us to detect invalid indexes and lets us tune how much work we are willing to do per car index.
+
+Full validation of every block is not recommended as it opens us up to performing unbounded work. *We have seen CAR files with millions of tiny blocks.*
```
vasco-santos, line 210: based on our previous chat, I was thinking the random validation would consider a % of the blocks in a CAR. But reading this now, looks like a specific number. Would that be the case? I would be more in favour of a random %, but probably a good idea to add a custom MAX. Otherwise, there is the attack vector of uploading gigantic CAR of tiny blocks to better try to not get bad ones validated

alanshaw: Yeah percent sounds good, but maybe with a max number we're willing to consider...in the spec I'd probably just specify a "random sample that may inculde none or all of the blocks" though.

w3-ipni.md
```diff
@@ -0,0 +1,237 @@
...
+
+- [olizilla], [Protocol Labs]
+
+# Abstract
+
+For IPNI we assert that we can provide batches of multihashes by signing "Advertisements". 
+
+With an inclusion claim, a user asserts that a CAR contains a given set of multihashes via a car index.
+
+This spec describes how to merge these two concepts by adding an `ipni/offer` capability to submit an inclusion claim as an IPNI Advertisement.
```
alanshaw, line 15: `ipni/offer` implies an `ipni/accept` fx per our own conventions...

It might be good to have a `ipni/accept` task that is invoked when the advert has been written. The receipt might include the advert (C)ID and an identifier for the chain that it is included in.

w3-ipni.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Introduction 
+
+**What this unlocks** (tl;dr)
+
+- Create 1 or more IPNI Adverts per user uploaded CAR and set the ContextID to be the CAR CID (instead of arbitrary batches with no ContextId)
```
alanshaw, line 25: Shard CID + Space DID?

alanshaw (commented):
ERHMAHGERD linter is not happy about this...

w3-ipni.md
olizilla: I think this should be Shard CID only. If we publish the same set of multihashes again to IPNI becuase someone adds the same CAR to a new space, I don't think we want to double the set of results that come back from IPNI for that multihash (1 per space it's in), all the records would have the same set of provider info (w3s), and theres no mechanism to determine which of the space dids the user should pass to us when reading (if at all).

related... I'm not sure what happens if we publish the same multihash multiple times to IPNI with different ContextIDs. I think you get multiple results back with same provider info.

w3-ipni.md
olizilla: I'm in to this from the angle: "lets expose our block level index info in a way that is easy to replicate rather than trapping it in a private db"... decentralise our indexes.

I don't think IPNI team would jump at the chance to host this but I'm in favour of seeing it as a separate service. I think we'll bundle it in to the client as part of the default upload flow to start with, but we can break it out and make it opt-in / hosted elsewhere once we have this in place.

w3-ipni.md
olizilla: i think this comment is intended for the validation section. This is the encode it as 1 or more adverts part which is just dealing with max block size.

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+```json
+{
+  "content": CID, // CAR CID
+  "includes": CID // CARv2 Index CID
+}
+```
+
+When `ipni/offer` is invoked the service must fetch the inclusion claim. The encoded claim block may be sent with the invocation.
+
+The service must fetch he CARv2 index and parse it to find the set of multihashes included in the CAR. see: [Verifying the CARv2 Index](#verifying-the-carv2-index)
```
Gozala, line 178: I'm bit confused here fetch from where ? Should not index be send along with the claim ?

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+    Entries Link
+    ContextID Bytes
+    Metadata Bytes
+    IsRm Bool
+    ExtendedProvider optional ExtendedProvider
+}
+```
+
+- `Entries` must be the CID of an `EntryChunk` for a subset (or all) of multihashes in the CAR.
+- `ContextID` must be the byte encoded form of the CAR CID.
```
Gozala, line 197:
```suggestion
- `ContextID` must be the byte encoded form of the CAR CID.
- `Entries` must be the CID of an `EntryChunk` for a subset (or all) of multihashes in the CAR.
```

Gozala: I think it's easier to follow if you mention what context is first as it's referenced from the other field.

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+- `Metadata` must be the bytes of the inclusion claim.
```
Gozala, line 198: It's a shame to loose the provenance info as in where the claim was originated from, would be nice to capture the source `ipni/offer` somewhere.

Gozala: Also I think it would be nice if `ipni/offer` had an `ipni/add` effect that would link to our IPNI advertisements, similar to what we do with filecoin. This would even allow user to notify indexer nodes without waiting on us to do it.

I am also very tempted to be storing advertisements in user space as opposed to our own custom bucket, if they delete it we can then publish delete advertisement.

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+
+![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+## Authors
+
+- [olizilla], [Protocol Labs]
+
+## Abstract
+
+For [IPNI] we assert that we can provide batches of multihashes by signing "Advertisements".
```
gobengo, line 11:
```suggestion
For [IPNI] we assert that we can provide batches of multihashes by signing ["Advertisements"](https://github.com/ipni/specs/blob/main/IPNI.md#advertisements).
```

gobengo: nvm I see it is linked later

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+
+### Verifying the CARv2 Index
+
+The service must fetch the CARv2 Index and may verify 1 or more multihashes from the index exist at the specified offsets in the associated CAR.
+
+The verifier should pick a set of multihashes at random and fetch the bytes from the CAR identified by the index entry and verify it's multihash. The invocation must return an error if any entry is found to be invalid.
+
+Random validation of a number of blocks allows us to detect invalid indexes and lets us tune how much work we are willing to do per car index.
+
+Full validation of every block is not recommended as it opens us up to performing unbounded work. _We have seen CAR files with millions of tiny blocks._
```
Gozala, line 212: I worry about the probability of considering index valid while in practice it is not the case. It also might be better to validate that random byte offsets in the CAR are correctly indexed as opposed to validating random blocks.

## 💡🤯

Could we actually make clients create advertisements signed with their own keypair and just our addresses so that we don't have to verify and simply enable users to publish to IPNI ?

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+### Encoding the IPNI Advertisement
+
+> The set of multihashes must be encoded as 1 or more [IPNI Advertisements].
+
+In IPNI, batches of multihashes are encoded as `EntryChunk` blocks, each batch includes an array of multihashes.
+
+A `MultihashIndexSorted` Index encodes a set of multihashes. Mapping from an index to an `EntryChunk` requires parsing the index and encoding the multihashes it contains with the EntryChunk IPLD schema.
+
+```ipldsch
+type EntryChunk struct {
```
Gozala, line 223: I find `EntryChunk` to be very misleading name, because it contains info about multiple entries not one. Can this be renamed to something less confusing like `EntryBatch` or `AdvertisedEntrySet` ?

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+    Entries [Bytes]
+    Next optional Link
+}
+```
+
+Where the IPLD encoded size of an `EntryChunk` with the set of multihashes would exceed 4MiB (the upper limit for a block that can be transferred by libp2p) the set of multihashes must be split into multiple `EntryChunk` blocks.
```
Gozala, line 229: I would recommend sorting all entries first before putting them into batches.

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+
+An `Advertisement` includes `Provider` info which claims that a the batch of multihashes are available via bitswap or HTTP, and are signed by the provider PeerID private key; Each advert is a claim that this peer will provide that batch of multihashes.
+
+Advertisements also include a CID link to any previous ones from the same provider forming a hash linked list.
+
+The latest `head` CID of the ad list can be broadcast over [gossipsub], to be replicated and indexed by all listeners, or via HTTP to specific IPNI servers as a notification to pull and index the latest ads from you at their earliest convenience.
+
+The advert `ContextID` allows providers to specify a custom grouping key for multiple adverts. You can update or remove multiple adverts by specifying the same `ContextID`. The value is an opaque byte array as far as IPNI is concerned, and is provided in the query response.
+
+A `Metadata` field is also available for provider specific retrieval hints, that a user should send to the provider when making a request for the block, but the mechanism here is unclear _(HTTP headers? bitswap?)_.
```
gobengo, line 87: I think it would be good to put a URL like `https://web3.storage/ipfs/{cid}` here. That alone is enough to allow http clients to e.g. do a HEAD request or even webfinger request to learn more about the supported representations of the resource and authorization required to retrieve it.

(Can also put a multiaddr in there but IMO it would be nice for it to be represented as a URI https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml e.g. `web+libp2p:{multiaddr}` but 🤷‍♀️)

Gozala (commented):
I think this is really cool! I really wish we captured the byte ranges however as that would allow reads without having to do extra lookups.

I also wish we would generalize this from the CARs. We have already started talking about writing files as is and just storing proof trees for them instead of transcoding things into CARs etc... In that future context is likely going to be root CID of the DAG and index will be byte ranges either in the source file or a proof. I think it would be a good idea to design protocol with that in mind instead and if source payload will happens to be a CAR right now that is fine also.

gobengo (commented):
It seems like the invocation to get an IPNI advertisement into web3.storage could just be an `upload/add` to a space?

And then if we had a `upload/read` nb.cid capability, you could delegate it to `"aud": "did:web:cid.contact"` and it could invoke it as desired to fetch the blocks? But then we'd also have receipts/billing on the read side of this, where it seems like it could be useful to have a different cost for reading the adverts 1 time every 2 weeks vs 1billion times per day, and furthermore the end-user may want to authorize advertisement reads from IPNI deployments other than cid.contact.

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+}
+```
+
+- `Entries` must be the CID of an `EntryChunk` for a subset (or all) of multihashes in the CAR.
+- `ContextID` must be the byte encoded form of the CAR CID.
+- `Metadata` must be the bytes of the inclusion claim.
+
+See: [Encoding the IPNI Advertisement](#encoding-the-ipni-advertisement)
+
+The Advertisement should then be available for consumption by indexer nodes per the [Advertisement Transfer](https://github.com/ipni/specs/blob/main/IPNI.md#advertisement-transfer) section of the IPNI spec.
```
gobengo, line 202: It seems the advertisement transfer section also has an affordance for submitting advertisements
> Announcements are sent as HTTP PUT requests to /announce on the indexer node's 'ingest' server.

So it seems like the client can already send cid.contact the Advertisements directly. How would dev decide whether to to do that or go through this `ipni/offer` proposal?

w3-ipni.md
olizilla: The Metadata field here is at the Advertisement level, so it's about an arbitrary batch of CIDs rather than one. We could potentially put the a gateway url for the carCID at that level. The proposal here is to use this space for the bytes of the inclusion claim.

In IPNI today, an Advertisement maps an EntryChunk CID (aka a CID for a batch of multihashes) to a Provider. A Provider is an array of multiaddrs, that define the location and transport^ to use to fetch them. A URL would be nice, but PL-ware is multiaddrs everywhere.

^But, gotcha! right now specifying a multiaddr with "http" as a transport isn't enough to say "pls use trustless ipfs gateway semantics when retreving over http", so the Metadata field is provided to give more hints that can't yet be inferred from the multiaddr alone... Bitswap has a similar issue... the multiaddr would specify "p2p" as the transport, but then libp2p protocol negotiation is left to the peers to decide on. The Metadata field again has a varint to declare Bitswap should be expected.

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+
+Advertisements also include a CID link to any previous ones from the same provider forming a hash linked list.
+
+The latest `head` CID of the ad list can be broadcast over [gossipsub], to be replicated and indexed by all listeners, or via HTTP to specific IPNI servers as a notification to pull and index the latest ads from you at their earliest convenience.
+
+The advert `ContextID` allows providers to specify a custom grouping key for multiple adverts. You can update or remove multiple adverts by specifying the same `ContextID`. The value is an opaque byte array as far as IPNI is concerned, and is provided in the query response.
+
+A `Metadata` field is also available for provider specific retrieval hints, that a user should send to the provider when making a request for the block, but the mechanism here is unclear _(HTTP headers? bitswap?)_.
+
+Regardless, it is space for provider specified bytes which we can use as to include the portable cryptographic proof that an end-user made the original claim that a set of blocks are included in a CAR and that as a large provider we have alerted IPNI on their behalf.
```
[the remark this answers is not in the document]
olizilla: IPNI Advertisements **must** be signed with the PeerID key that should be used to secure the lib2p connection. In the case of Bitswap we don't have a choice. We must sign with the same key we use for the libp2p node that we, the provider, is claiming we can provide.

This for me is _the_ interesting integration point with IPNI. Our user-provided content claims are assertions from content creators about the content. IPNI is assertions from providers about what can be fetched from them.

Including the user provided content claim in the IPNI advert at least allows the possibility for consumers to maintain a reputation system that considers the original claimants key rather than attributing all indexing errors to the large upload service.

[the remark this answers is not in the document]
olizilla: however! for trustless-ipfs-gateway flavour http, the peerID key isn't used the secure the connection. it's http, the tls cert is used instead. so there is no requirement for the signing key for trustless http flavour providers to be anything in particular.

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+
+### How web3.storage integrates IPNI today
+
+web3.storage publishes IPNI advertisements as a side-effect of the E-IPFS car [indexer-lambda].
+
+Each multihash in a CAR is sent to an SQS queue. The `publisher-lambda` takes batches from the queue, encodes and signs `Advertisement`s and writes them to a bucket as JSON.
```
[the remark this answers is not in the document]
olizilla: it writes a `head` file to s3 for durability, and we have to configure the sqs consumer to only allow a single instance of the lambda to run at once. It's not great.

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+
+The lambda makes an HTTP request to the IPNI server at `cid.contact` to inform it when the head CID of the Advertisement linked list changes.
```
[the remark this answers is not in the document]
olizilla: No, it uses HTTP. https://github.com/ipni/specs/blob/main/IPNI.md#advertisement-transfer

We tell it the CID of the new head, we write files with the CID as the filename. There is also a pre agreed `/head` file which ingest servers can poll to see if the head has changed, and if we are still "up". If we stop responding to `/head` requests, then at some point the IPNI server may flag our content / drop it from search results / delete it from their db.

w3-ipni.md
olizilla: IPNI Advertisements have to be signed with the PeerID key that should be used to secure the connection when trying to fetch the bytes from the provider. As such an end user can't tell IPNI that web3.storage will provide these blocks, as they can't sign it on our behalf.

This is an issue for libp2p based connections where the PeerID is used to authenticate and secure the connection, but not for trustless ipfs gateway flavour http, where tls is used and there is no notion of a PeerID.

w3-ipni.md
olizilla: No. This is what we all want, but is not an option for libp2p based protocols where the IPNI Advert signer and the provider PeerID must be the same.

We could explore it for trustless ipfs gateway providers like w3s.link, but, I'd rather explore a solution that works for both first.

w3-ipni.md
olizilla: Not in this spec. This is the IPNI vocabulary. https://github.com/ipni/specs/blob/main/IPNI.md#entrychunk-chain

w3-ipni.md
gammazero: Transforming multiaddr to URL is simple enough. Will the client be querying IPNI directly, or will some lookup service be doing that and then generating a possibly signed URL that the client can use to get the data?

w3-ipni.md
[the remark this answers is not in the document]
gammazero: > Including the user provided content claim in the IPNI advert

It may not be practical to put claims in the IPNI advertisement metadata if they require much data to store. It may be necessary to put a reference (CID or IPNS name) to the associated claims into the metadata, and fetch claims  separately. If an IPNS name is used then the claims can be changed without changing the reference stored in the metadata.

w3-ipni.md
gammazero: Instead of the metadata being the bytes of the inclusion claim, what about having the metadata contain a reference to an external body of data (CAR file?) that contains a set of claims and possibly other information? This would require an extra fetch. This external (not in IPNI) data could also contain any items needed to grant access to specific clients

w3-ipni.md
gammazero: > Could we actually make clients create advertisements signed with their own keypair and just our addresses

IPNI does allow for an advertisement publisher to publish (and sign) ads on behalf of a provider. However, this creates other issues around maintaining and making/keeping the advertisement chains available to indexers.

w3-ipni.md
gammazero: It is baked into the IPNI spec and encoded into existing advertisements. An `EntryChunk` can contain information about only one multihash, but usually contains more.

`AdvertisedEntrySet` could also be misleading, since all the multihashes advertised by an advertisement are part of the advertised set. `EntryBatch` seems better, but does not indicate that individual batches are related or part of a large unit, which they absolutely are if they are associated with the same advertisement.

The term `EntryChunk` was chosen because it serves the same purpose as [HTTP chunking](https://en.wikipedia.org/wiki/Chunked_transfer_encoding), and the term "chunk" implies that it is part of a larger collection. There may be better names, but IMO this is fairly accurate.

w3-ipni.md
gammazero: Why?

Depending on the storage backend within IPNI, these do get sorted for sharding or index-based merging, etc.

w3-ipni.md
gammazero: @gobengo The section you cite above:
> Announcements are sent as HTTP PUT

Talks about `Announcements` not `Advertisements`. These are different things. An [announcement](https://pkg.go.dev/github.com/ipni/go-libipni@v0.6.2/announce#Announce) tells IPNI that a new advertisement is available at a particular address and the peer ID of the host for that advertisement. But, yes, a client could send the announce message to IPNI if it knows that data.

w3-ipni.md
gammazero: I suggest a configurable percentage (validation factor) from 0 to 100%. Any non-zero fraction of a number of blocks is rounded up to the nearest integer. So that 10 blocks at 3% validation, still validates one block.

w3-ipni.md
gammazero: Who is the consumer of that receipt? This would be a copy of the `Announce` message, that has the advertisement CID and the peerID and addresses of where the chain is available from. The peerID (publisher ID) would identify which chain the ad is on.

w3-ipni.md
```diff
@@ -0,0 +1,246 @@
...
+
+[web3.storage] publishes the blocks it can provide by encoding a batch of multihashes as an IPLD object and writing it to a bucket as an `Advertisement`, addressed by it's CID.
+
+An `Advertisement` includes `Provider` info which claims that the batch of multihashes are available via bitswap or HTTP, and are signed by the providers PeerID private key; Each advert is a claim that this peer will provide that batch of multihashes.
+
+Advertisements also include a CID link to any previous ones from the same provider forming a hash linked list.
+
+The latest `head` CID of the advert list can be broadcast over [gossipsub], to be replicated and indexed by all listeners, or sent via HTTP to specific IPNI servers as a notification to pull and index the latest ads from you at their earliest convenience.
+
+The advert `ContextID` allows providers to specify a custom grouping key for multiple adverts. You can update or remove multiple adverts by specifying the same `ContextID`. The value is an opaque byte array as far as IPNI is concerned, and is provided in the query response.
```
gammazero, line 87: The `ContextID` also serves as a key that refers to specific metadata, and is used to update or delete that metadata. Updating metadata changes the metadata returned by IPNI lookups for all CIDs that were advertised with that context ID.

w3-ipni.md
```diff
@@ -0,0 +1,246 @@
...
+
+A `Metadata` field is also available for provider specific retrieval hints, that a user should send to the provider when making a request for the block, but the mechanism here is unclear _(HTTP headers? bitswap?)_.
+
+Regardless, it is a field we can use to include the portable cryptographic proof of the content-claim that an end-user made that a set of blocks are included in a CAR. The provider has to sign the IPNI advert with the peerID key that should be used to secure the libp2p connection when retrieving the block. For upload services like web3.storage,
```
gammazero, line 91: The functionality that is needed here is for the data owner to assert that the advertised CIDs are in the CAR file that is referred to by the context ID. In other words, this advertisement is correct for the CAR file. The advertisement is signed by the provider of the advertisement (w3s?), so the functionality here is adding the data owner's signature to the advertisement.

The inclusion claim carried by the metadata is limited to specifying a CAR index (by CID) that is associated with  the CAR CID

w3-ipni.md
gammazero: No. The claim is returned from IPNI and only associates a CARv2 index CID with the CAR CID. The CARv2 index will generally be far too big to store in indexer metadata. So, the CARv2 index must be fetched separately.

w3-ipni.md
```diff
@@ -0,0 +1,242 @@
...
+**Inclusion claim** example
+
+```json
+{
+  "content": CID, // CAR CID
+  "includes": CID // CARv2 Index CID
+}
+```
+
+When `ipni/offer` is invoked the service must fetch the inclusion claim. The encoded claim block may be sent with the invocation.
```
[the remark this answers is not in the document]
gammazero:
```suggestion
When `ipni/offer` is invoked, the encoded inclusion claim block is sent with the invocation.
```

w3-ipni.md
gammazero: I like the idea of storing advertisements in user space. That way the user pays to index the advertisements as part of the storage cost. We will need to generate events when a file is deleted to that a removal advertisement can be created.

Can a user opt-out of indexing?

w3-ipni.md
```diff
@@ -0,0 +1,246 @@
...
+```ipldsch
+type EntryChunk struct {
+    Entries [Bytes]
+    Next optional Link
+}
+```
+
+Where the IPLD encoded size of an `EntryChunk` with the set of multihashes would exceed 4MiB (the upper limit for a block that can be transferred by libp2p) the set of multihashes must be split into multiple `EntryChunk` blocks.
+
+It is possible to create long chains of `EntryChunk` blocks by setting the `Next` field to the CID to another `EntryChunk`, but this requires an entire EntryChunk to be fetched and decoded, before the IPNI server can determine the next chunk to fetch.
```
gammazero, line 233: This is not a problem since indexing is not guaranteed to be immediate, and it is much faster than having the same multihashes split over multiple advertisements.

w3-ipni.md
```diff
@@ -0,0 +1,246 @@
...
+
+The containing CAR CID provides a useful `ContextID` for grouping multiple (light weight) Advertisement blocks so it is recommended to split the set across multiple `Advertisement` blocks each pointing to an `EntryChunk` with a partition of the set of multihashes in, and the `ContextID` set to the CAR CID.
```
gammazero, line 235: > it is recommended to split the set across multiple `Advertisement` blocks

I do not understand or agree with this recommendation since reading the same data from several advertisements will be much slower and require much more data transfer than reading from multiple entries blocks in the same advertisement.

Advertisements are also chained together with a field that points to the previous advertisement in the chain, so they also need to be decoded and read sequentially. Advertisements carry much more information (context ID, metadata, provider info, signature, etc.), whereas entries blocks contain only the multihashes and a link to the nexd block.

w3-ipni.md
```diff
@@ -0,0 +1,246 @@
...
+    Metadata Bytes
+    IsRm Bool
+    ExtendedProvider optional ExtendedProvider
+}
+```
+
+- `Entries` must be the CID of an `EntryChunk` for a subset (or all) of multihashes in the CAR.
+- `ContextID` must be the byte encoded form of the CAR CID.
+- `Metadata` must be the bytes of the inclusion claim.
+
```
gammazero, line 201: Do we want to include where the content claims for this CAR can be retrieved from?
`- Addresses` endpoint addresses where content claims can be retrieved from.

The content claims will include location claims, and there will be a location claim for the CAR file and the CARv2 index.

Or should these location claims be encoded into the metadata with an additional location for the location of all the content claims?

w3-ipni.md
Gozala: General direction I'm advocating for is that if you upload content to our service it just sits there without been indexed or advertised anywhere. If user wants to make it readable they must issue invocation requesting a location claim to be made for the uploaded content, which will in turn index and advertise.

> We will need to generate events when a file is deleted to that a removal advertisement can be created.

Deletes happen on user invocation which can be a trigger to remove an advertisement.

w3-ipni.md
Gozala: When I keep say derive keys from user identifier I imply that we could derive private key on demand and consequently could have those synthetic peer's, but I think I'm getting into the weeds here.

My main sentiment is that currently we are incentivized to verify claims made by users otherwise our reputation might take hit. This is not a great, ideally we would allow users to advertise claims and let their reputation suffer if those claims appear invalid.

There are possibly other ways to accomplish this e.g produce synthetic advertisement keys for `(user_did, cid)` tuples, but unless there is a way to query by `(*, cid)` it will likely not be very practical.

w3-ipni.md
Gozala: clients currently don't send CAR indexes.

PR #86, opened 2023-12-18
feat: proposal for content claims

alanshaw:
I did the minimal amount of work to move https://hackmd.io/IiKMDqoaSM61TjybSxwHog?view here 😄 

I did not include anything after "implementation notes" in the original.

### DIFF
--- a/content-claims.md
+++ b/content-claims.md
@@ -1,0 +1,334 @@
+# Content Claims Protocol
+
+![wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+## Editors
+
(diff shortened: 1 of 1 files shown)


content-claims.md
```diff
@@ -0,0 +1,334 @@
...
+  * GET Body MUST match multihash digest in CID.
+* Using CAR CID, or any future BlockSet() identifier, allows this to be a multi-block interface.
+
+```javascript
+{
+  "op": "assert/location",
+  "rsc": "https://web3.storage",
+  "input": {
+    "content" : CID /* CAR CID */, 
+    "location": "https://r2.cf/bag...car",
```
vasco-santos, line 90: note that implementation currently receives an array here, which I would think is also the best option :)

PR #92, opened 2024-01-18
fix: add filecoin info capability

vasco-santos:
Adds previously implemented `filecoin/info` capability. Also changes spec state to reflect current state

### DIFF
--- a/w3-filecoin.md
+++ b/w3-filecoin.md
@@ -1,6 +1,6 @@
 # W3 Filecoin Protocol
 
-![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+![status:reliable](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square)
 
 ## Editors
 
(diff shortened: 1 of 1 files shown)


w3-filecoin.md
```diff
@@ -363,6 +365,96 @@ If the added piece is invalid, the reason for the failure is also reported:
...
+
+> `did:key:zAliceAgent` invokes `filecoin/info` capability provided by `did:web:web3.storage`
+
+```json
+{
+  "iss": "did:key:zAliceAgent",
+  "aud": "did:web:web3.storage",
+  "att": [
+    {
+      "with": "did:key:zAlice",
```
alanshaw, line 380: In the client we're using the space DID.

vasco-santos: this is what it was meaning to reflect here. We have `"did:key:zAliceAgent"` as the agent and `"did:key:zAlice"` as the space DID across all Storefront capabilities.

Should I add Space suffix to the resource of all `filecoin/*` to be more clear?

alanshaw: Oh lol ok!

alanshaw (approved):

PR #93, opened 2024-01-18
docs: add w3 space basic spec

vasco-santos:
Adds basic space spec. While at first considered to add it to `w3-account` or `w3-access`, I realized that account has a specific focus on high level account with multiple spaces, while access focus on "aggregating and managing delegated capabilities across agents". Therefore, ended up with new spec document to reflect the basics of a space protocol.

It is complemented by `w3-account` and `w3-provider` on how to integrate account and providers, while in its basic form adds wire protocol for the space specific capabilities that do not fall under the umbrella of `account` and `provider`.

### DIFF
--- a/w3-space.md
+++ b/w3-space.md
@@ -16,7 +16,6 @@ A Space can be defined as a namespace for stored content. It is created locally,
 
 - [Capabilities](#capabilities)
   - [`space/info`](#spaceinfo)
-  - [`space/allocate`](#spaceallocate)
 
 ## Language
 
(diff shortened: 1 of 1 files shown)


w3-space.md
```diff
@@ -0,0 +1,126 @@
...
+      "did": "did:key:zAlice",
+      "providers": [
+        "did:web:free.web3.storage"
+      ]
+    }
+  }
+}
+```
+
+### `space/allocate`
```
alanshaw, line 77: Do we want to document this? AFAIK it is not used by our client/server as an invoked capability...although the handler _is_ used by `store/add` handler.

However it doesn't do anything related to allocating space - it rate limits and checks for storage provider.

alanshaw (changes requested):

w3-space.md
vasco-santos: it is used by `store/add` yes. I think it should be documented, as `store/add` may happen in another place in the future. 

> However it doesn't do anything related to allocating space - it rate limits and checks for storage provider.

I think because it is not implemented? But that is the goal here? We can make that clear in the spec if that would be the desire.

Maybe @travis can help out with decision here

alanshaw: > it is used by `store/add` yes. I think it should be documented, as `store/add` may happen in another place in the future.

Right but I think that should be in `w3-store.md` and should say that `store/add` MAY rate limit and SHOULD ensure the space is provisioned properly.

> I think because it is not implemented? But that is the goal here? We can make that clear in the spec if that would be the desire.

I would just remove it - it is not invoked and does nothing to do with allocating space. The handler function should instead be a lib/utility function.

vasco-santos: ok, I removed until we decide what to do

alanshaw (approved):

PR #94, opened 2024-01-18
docs: add w3 plan basic spec

vasco-santos:
Adds basic specs for `plan/*` protocol

### DIFF
--- a/w3-plan.md
+++ b/w3-plan.md
@@ -12,7 +12,7 @@
 
 ## Abstract
 
-A Plan
+A Plan that accounts can register to, in order to have an associated billing plan for their spaces.
 
 - [Capabilities](#capabilities)
   - [`plan/get`](#planget)
--- a/w3-plan.md
+++ b/w3-plan.md
@@ -1,0 +1,121 @@
+# Space
+
+![status:reliable](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square)
+
+## Editors
+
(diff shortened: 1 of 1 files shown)


w3-plan.md
```diff
@@ -0,0 +1,121 @@
...
+
+- [Vasco Santos], [Protocol Labs]
+
+## Authors
+
+- [Vasco Santos], [Protocol Labs]
+
+## Abstract
+
+A Plan
```
alanshaw, line 15: Can expand on this 🙏 ?

alanshaw (changes requested):

w3-plan.md
vasco-santos: oh yeah, failed to commit that 🤦🏼

w3-plan.md
```diff
@@ -0,0 +1,121 @@
...
+
+- [Vasco Santos], [Protocol Labs]
+
+## Authors
+
+- [Vasco Santos], [Protocol Labs]
+
+## Abstract
+
+A Plan that accounts can register to, in order to have an associated billing plan for their spaces.
```
alanshaw, line 15: Hmm, ok how about:

```suggestion
A plan specifies limits and restrictions on usage. Different plans have different limits and different price points. Accounts select a plan so that they can be billed the correct amount each period.
```

alanshaw (changes requested):

alanshaw (approved):

PR #97, opened 2024-01-19
add implementations for w3-account.md

gobengo:
Motivation:
* add links to known implementations so people seeking more context about how the protocol can be used can look to how they are implemented

### DIFF
--- a/w3-account.md
+++ b/w3-account.md
@@ -1,6 +1,6 @@
 # Account
 
-![status=reliable](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square)
+![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
 
 ## Editors
 
--- a/w3-account.md
+++ b/w3-account.md
@@ -147,31 +147,31 @@ The authorization session signature is denoted by a [Nonstandard `VarSig` signat
 
 ### [`w3 login <email>` in w3cli](https://github.com/web3-storage/w3cli#w3-login-email)
 
-* invokes [Account.login](https://github.com/web3-storage/w3cli/blob/fc97ee1b76551bced861f08a4d1e7a31440a6a14/bin.js#L56) which calls `login` on a `@web3-storage/w3up-client`
+- invokes [Account.login](https://github.com/web3-storage/w3cli/blob/fc97ee1b76551bced861f08a4d1e7a31440a6a14/bin.js#L56) which calls `login` on a `@web3-storage/w3up-client`
 
 ### [@web3-storage/w3up-client]()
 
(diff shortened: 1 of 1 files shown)


w3-account.md
```diff
@@ -1,6 +1,6 @@
 # Account
 
-![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+![status=reliable](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square)
```
gobengo, line 3:
```suggestion
![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
```

w3-account.md
```diff
@@ -25,7 +25,9 @@ In w3 family of protocols, a namespace is identified by a [`did:key`] identifier
 1. Synchronizing delegations to namespaces across multiple user agents on multiple devices is difficult because of the use of non-memorable [`did:key`] identifiers.
 2. Recovering access if the user loses access to all devices is also a challenge.
 
-To address these issues, we propose the concept of an account as a way to aggregate and manage capabilities under a human-meaningful identifier such as an email address. Specifically, we propose deriving an account identifier from a user-controlled email address in the form of a [`did:mailto`] identifier, which can act as the [principal] in [UCAN] delegation chains.
+To address these issues, we propose the concept of an account as a way to aggregate and manage capabilities under a human-meaningful identifier such as an email address.
+
+Specifically, with did:mailto Accounts we propose deriving an account identifier from a user-controlled email address in the form of a [`did:mailto`] identifier, which can act as the [principal] in [UCAN] delegation chains.
```
alanshaw, line 30: I think the previous version was clearer. I read it as this document is proposing an account that is identified by an email address, in the form of a `did:mailto` DID.

In your change you've added `did:mailto` _before_ it is introduced...

```suggestion
Specifically, we propose deriving an account identifier from a user-controlled email address in the form of a [`did:mailto`] identifier, which can act as the [principal] in [UCAN] delegation chains.
```

alanshaw (changes requested):

w3-account.md
gobengo: @alanshaw I'm trying to introduce a distinction between the two ideas:
* " we propose the concept of an account as a way to aggregate and manage capabilities under a human-meaningful identifier such as an email address."
* and the more specific idea of using an email address as the identifier

i.e. was thinking the former could be 'accounts' and the latter more specifically 'did:mailto accounts'  (accounts that uses did:mailto as opposed to another identifier).

`did:mailto` is somewhat useful for specific use cases, but I'd love to be able to explain that you can use other DIDs too and not just that one.

Just explaining my rationale. It's not super important for me to update this copy right now, and not obvious how to do so for the above messaging goals. So i'm going to accept your suggestion here. (I'm also happy to un-modify this section alltogether since it is irrelevant to addign implementations, my bad)

PR #102, opened 2024-01-20
feat: ucan ext

Gozala:
Document UCAN spec extensions we use and assume. Based on #31


w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+| Name        | Description |
+| ----------- | ----------- |
+| Principal | The general class of entities that interact with a UCAN. Listed in the `iss` or `aud` field |
+| Issuer | Principal sharing access. It is the signer of the [UCAN]. Listed in the `iss` field |
+| Audience | Principal access is shared with. Listed in the `aud` field |
+
+### Authority
+
+Authority is trusted [DID] identifier. For example various subsystems may recognize and signatures from a global service authority.
```
alanshaw, line 35:
```suggestion
Authority is trusted [DID] identifier. For example various subsystems may recognize signatures from a global service authority.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+Subsystem that performs [UCAN] validation.
+
+# Capabilities
+
+## Attestation
+
+### Motivation
+
+[UCAN] verification process involves going through a delegation chain and verifying that every delegation has a valid signature from the private key if the issuer and that principals in the chain align. As described it has following implications:
```
alanshaw, line 49:
```suggestion
The [UCAN] verification process involves going through a delegation chain and verifying that every delegation has a valid signature from the private key of the issuer and that the principals in the chain align. As described it has the following implications:
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+1. Verifier MAY end up verifying same potentially long chain over and over. As an optimization they could implement caching strategy, but that may require access to some storage.
+1. Invocations MAY have to send potentially long delegation chains over and over as validator MAY prune cache and may not even persist previously seen delegations.
+1. Verifying delegations issued by principals other than [`did:key`] involves some, potentially expensive, [DID] document resolution. Mutable nature of [DID] document also imply that previously valid delegation chain could be rendered invalid.
+
+### Proposal
+
+We propose extension to the core [UCAN] specification and define `ucan/attest`
```
alanshaw, line 57:
```suggestion
We propose an extension to the core [UCAN] specification and define a `ucan/attest`
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+capability, that enables an [authority] to attest that linked [UCAN] delegation is valid.
```
alanshaw, line 58:
```suggestion
capability, that enables an [authority] to attest that a linked [UCAN] delegation is valid.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+This effectively allows distributing verifiable cache records to interested principals so they could include those in subsequent invocations and take advantage of optimizations they unlock.
```
alanshaw, line 60:
```suggestion
This effectively allows distributing verifiable cache records to interested principals so they can include those in subsequent invocations and take advantage of the optimizations they unlock.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### Cached Verification
+
+For example verifier could issue an attestation for the [UCAN] chain it validated and hand it back to the caller. In subsequent calls caller could include an attestation to take advantage of more optimal execution.
```
alanshaw, line 64:
```suggestion
For example, a verifier could issue an attestation for the [UCAN] chain it validated and hand it back to the caller. In subsequent calls the caller could include an attestation to take advantage of more optimal execution.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### Proof chain compaction
+
+Just like in the above scenario except not only caller can take advantage of more optimal execution they could also transfer proof chain up to a proof they have attestation for reducing size of the payload.
```
alanshaw, line 68:
```suggestion
Just like in the above scenario except not only can a caller take advantage of more optimal execution, they can also transfer a proof chain up to a proof they have an attestation for, thereby reducing the size of the payload.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+#### Out of bound verification
+
+Attestations CAN be issued as a part of authorization process. For example [UCAN]s issued by [`did:mailto`] require out of bound authorization flow in which service sends confirmation email to the user. If user confirms through a link in the email service issues an attestation declaring that [UCAN] has been verified.
```
alanshaw, line 72:
```suggestion
Attestations CAN be issued as a part of authorization process. For example [UCAN]s issued by [`did:mailto`] require out of bound authorization flow in which a service sends a confirmation email to the user. If the user confirms through a link in the email, the service issues an attestation declaring that the [UCAN] has been verified.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+- [Irakli Gozalishvili], [Protocol Labs]
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+## Abstract
+
+[UCAN] specification reserves `ucan/` ability namespace for the core functionality that may be introduced in future versions. Here we define W3 protocol extensions to the core namespace that we hope to standardize in core [UCAN] specification.
```
alanshaw, line 19:
```suggestion
The [UCAN] specification reserves the `ucan/` ability namespace for the core functionality that may be introduced in future versions. Here we define W3 protocol extensions to the core namespace that we hope to standardize in core [UCAN] specification.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+Just like in the above scenario except not only caller can take advantage of more optimal execution they could also transfer proof chain up to a proof they have attestation for reducing size of the payload.
+
+#### Out of bound verification
+
+Attestations CAN be issued as a part of authorization process. For example [UCAN]s issued by [`did:mailto`] require out of bound authorization flow in which service sends confirmation email to the user. If user confirms through a link in the email service issues an attestation declaring that [UCAN] has been verified.
+
+### Notes on compatibility
+
+Attestations effectively are a cache records that if provided enables certain optimizations, when not provided, correct implementation SHOULD fallback to a less optimal path.
```
alanshaw, line 76:
```suggestion
Attestations are effectively cached records that, if provided, enable certain optimizations. When not provided, correct implementations SHOULD fallback to a less optimal path.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+Implementations that do not support this extension will simply disregard attestations in the proof chain as irrelevant and just under less optimal path.
```
alanshaw, line 78:
```suggestion
Implementations that do not support this extension will simply disregard attestations in the proof chain as irrelevant and just take a less optimal path.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+The value of the `with` field MUST be the DID of the [authority], indicating that ONLY [authority] authorized principals are able to issue attestations on behalf of the [authority].
+
+### Attestation Subject
+
+The `nb.proof` field MUST be a [link] to the [UCAN] delegation covered by the attestation.
+
+### Attestation Lifetime
+
+Attestation MUST be considered valid ONLY within the [time bounds] of enclosing [UCAN]. In other words enclosing [UCAN] `nbf` and `exp` fields SHOULD be used to specify [time bounds] within which attestation is valid.
```
alanshaw, line 109:
```suggestion
The attestation MUST be considered valid ONLY within the [time bounds] of the enclosing [UCAN]. In other words the enclosing [UCAN]'s `nbf` and `exp` fields SHOULD be used to specify [time bounds] within which the attestation is valid.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Attestation Revocation
+
+Attestation MUST be considered revoked if enclosing [UCAN] has been revoked. This also implies that attestation can be revoked by revoking their enclosing [UCAN]s.
```
alanshaw, line 113:
```suggestion
The attestation MUST be considered revoked if the enclosing [UCAN] has been revoked. This also implies that the attestation can be revoked by revoking its enclosing [UCAN]s.
```

w3-ucan.md
```diff
@@ -0,0 +1,213 @@
...
+    }
+  }
+}
+```
+
+## Revocation
+
+### Motivation
+
+[UCAN] specification defines [revocation record] that MUST be signed by the issuer in the delegation chain of the [UCAN] been revoked. From real world experience we find that requirement unfortunate. In plenty of cases various actors are delegated specific subset of capabilities from the supervisor to carry out their work. Supervisor MAY want give specific actor to revoke some capabilities on its behalf without giving it ability to invoke them and [revocation record] does not allow that.
```
alanshaw, line 146: I can't understand that last sentence 😟 ...

alanshaw (changes requested):
I'm not sure it's necessary to link to UCAN/DID etc. for every mention...

w3-ucan.md
Gozala, line 146:
```suggestion
[UCAN] specification defines [revocation record] that MUST be signed by the issuer in the delegation chain of the [UCAN] been revoked. From real world experience we find this requirement problematic, it is common to have a primary authority that delegates subset of their capability to various actors based on their role. In such setup it is often desired to have auditors that can revoke capabilities from misbehaving actors without been in the delegation chain.

In other words it is desired to have ability to grant revocation power to some actor without granting them invocation power.
```

PR #105, opened 2024-01-22
chore: remove ucan/revoke revocation section

Gozala:
@alanshaw proposed removing this section in https://github.com/web3-storage/specs/pull/102#discussion_r1462148503 although I was not sure why. I decided to keep it and create this PR so we can decide what to do with it in a followup.


alanshaw:
I thought it was an accidental repeat of the previous section. Please leave if not!

PR #112, opened 2024-02-26
feat: add specification for UCAN HTTP bridge

travis:
To support users in languages that do not have existing UCAN invocation implementations, we are going to launch a bridge that allows them to make simple HTTP requests with JSON bodies that we transform into proper UCAN invocations.

### DIFF
--- a/w3-ucan-bridge.md
+++ b/w3-ucan-bridge.md
@@ -43,7 +43,7 @@ Content-Type: application/json
 }
 ```
 
-And receive a dag-json-encoded list of UCAN bridge receipts like:
+And receive a DAG-JSON-encoded list of UCAN bridge receipts like:
 
 ```json
 [
@@ -133,7 +133,7 @@ async function deriveSigner(headerValue: string): Promise<ed25519.EdSigner> {
 }
 ```
 
-`Authorization` is an IPLD CAR [ed: unixfs encoded? need to understand Delegation archive format better] representing a UCAN delegation.
+`Authorization` is an IPLD CAR representing a UCAN delegation.
 It should grant the principal identified by `X-Auth-Secret` appropriate capabilities
 on the resource identified in the JSON body of the HTTP request.
 
--- a/w3-ucan-bridge.md
+++ b/w3-ucan-bridge.md
@@ -24,11 +24,12 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 A user can submit an HTTP request like (simplified for clarity):
 
-```
+```HTTP
 POST /bridge
 X-Auth-Secret: uNGUyOTA2OTRlYjNlZDJjNjE3ZTRkNzBlYzJiN2RkYTM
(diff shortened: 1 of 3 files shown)


w3-ucan-bridge.md
```diff
@@ -0,0 +1,174 @@
...
+
+![stable](https://img.shields.io/badge/status-stable-brightgreen.svg?style=flat-square)
+
+## Editors
+
+- [Travis Vachon](https://github.com/travis), [DAG House](https://dag.house/)
+
+## Authors
+
+- [Travis Vachon](https://github.com/travis), [DAG House](https://dag.house/)
```
alanshaw, line 11: Perhaps credit @Gozala here also?

w3-ucan-bridge.md
```diff
@@ -0,0 +1,174 @@
...
+
+# Abstract
+
+We have implemented a "bridge" that allows w3up users to interact with the service without implementing the UCAN invocation wire protocols.
```
alanshaw, line 15: I'd rephrase this as what this specification describes and why it exists.

w3-ucan-bridge.md
```diff
@@ -0,0 +1,174 @@
...
+Content-Type: application/json
+
+{
+  tasks: [
+    ["store/add", "did:key:z6Mkm5qHN9g9NQSGbBfL7iGp9sexdssioT4CzyVap9ATqGqX", {
+        "link": "bagbaierah5sr5zt3tqgkrixptqzyerpxp5vwyjlx3n5frp2tbnr3clqrmrqa",
```
alanshaw, line 36: Should be a `dag-json` encoded CID?

w3-ucan-bridge.md
```diff
@@ -0,0 +1,174 @@
...
+Content-Type: application/json
+
+{
+  tasks: [
+    ["store/add", "did:key:z6Mkm5qHN9g9NQSGbBfL7iGp9sexdssioT4CzyVap9ATqGqX", {
```
alanshaw, line 35: Why is this an array and not a subset of the [invocation payload](https://github.com/ucan-wg/invocation/blob/c596f04bbaf56c58df76bb578c6dfd53bb88bece/README.md#invocation-payload)?

w3-ucan-bridge.md
```diff
@@ -0,0 +1,174 @@
...
+    "sig": {
+      "/": {
+        "bytes": "7aEDQCiJYmYe9Gf25hv84NVN/fjN+udnT4Q65kVHFmQb1MPB2EmwDnR+S/TeYNkMxBdwIsNOKDwyTCqKOxvUhbWkGA8"
+      }
+    }
+  }
+]
+```
+
+A UCAN bridge receipt has two fields, `data` and `sig`:
```
alanshaw, line 104: What if it was `p` and `s` (payload & signature) like https://github.com/ucan-wg/invocation/blob/c596f04bbaf56c58df76bb578c6dfd53bb88bece/README.md#receipt-response

w3-ucan-bridge.md
```diff
@@ -0,0 +1,174 @@
...
+
+`data` contains the result of an invocation (in the nested `out` field) along with information about the invocation.
+`sig` is a signature over a CBOR-encoded version of the value of the `data` field.
+
+## Authorization
+
+The `X-Auth-Secret` and `Authorization` header values can be generated with the `bridge generate-tokens` command of `w3cli`:
```
alanshaw, line 111: Before tuition on how to generate `X-Auth-Secret`, I'd like to know:

* What it is
* Why it is necessary
* How it is encoded

w3-ucan-bridge.md
```diff
@@ -0,0 +1,174 @@
...
+}
+```
+
+`Authorization` is an IPLD CAR representing a UCAN delegation.
+It should grant the principal identified by `X-Auth-Secret` appropriate capabilities
+on the resource identified in the JSON body of the HTTP request.
+
+## Request Format
+
+The request body should be a JSON or CBOR-encoded map with one key named `tasks` whose value is an array of "task specifications" of the form:
```
alanshaw, line 140: `dag-json` or `dag-cbor` right? Personally I'd specify that everything must be `dag-json` and leave as an extension in the future to be able to send `dag-cbor` using HTTP `Content-Type` and `Accept` headers to identify request body and specify response.

w3-ucan-bridge.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+`X-Auth-Secret` is a base64url-multibase-encoded Uint8Array of arbitrary length that will be used to derive an ed25519 principal as follows:
+
+```typescript
+import { sha256 } from '@ucanto/core'
+import { ed25519 } from '@ucanto/principal'
+
+async function deriveSigner(headerValue: string): Promise<ed25519.EdSigner> {
+  const secret = base64url.decode(headerValue)
```
alanshaw, line 128: Add import for this please.

alanshaw (changes requested):

w3-ucan-bridge.md
travis: ah yea - @Gozala and I came up with this in a design session last week to avoid having to choose a version of the invocation spec to follow - I'm definitely open to alternate task syntax here in the future, but this is a nice convenience for the simplest cases and avoids having to pick a version of the invocation spec while it's still cooking!

w3-ucan-bridge.md
travis: ah that's a great idea - I'll make the change here and in the code, thanks!

w3-ucan-bridge.md
```diff
@@ -0,0 +1,193 @@
...
+    "sig": {
+      "/": {
+        "bytes": "7aEDQCiJYmYe9Gf25hv84NVN/fjN+udnT4Q65kVHFmQb1MPB2EmwDnR+S/TeYNkMxBdwIsNOKDwyTCqKOxvUhbWkGA8"
+      }
+    }
+  }
+]
+```
+
+A UCAN bridge receipt has two fields, `data` and `sig`:
```
alanshaw, line 113: Needs to be changed to `d` and `s` as you did below.

alanshaw (approved):
I approved anyway but you just need to update the naming for receipt fields earlier in the document.

PR #114, opened 2024-03-19
feat: revamp store & upload specs

Gozala:
Updates store and upload specs so they are more like other specs as opposed to been more like documentation.

### DIFF
--- a/w3-store.md
+++ b/w3-store.md
@@ -172,15 +172,18 @@ type StoreAddSuccess union {
 }
 
 type StoreAddDone struct {
-  with    SpaceDID
-  link    &ContentArchive
+  with        SpaceDID
+  link        &ContentArchive
+  allocated   Int
 }
 
 type StoreAddPending struct {
-  with      SpaceDID
-  link      &ContentArchive
-  url       URL
-  headers   HTTPHeaders
+  with        SpaceDID
+  link        &ContentArchive
+  url         URL
+  headers     HTTPHeaders
+
+  allocated   Int
 }
 
 type URL = string
@@ -195,6 +198,8 @@ type StoreAddFailure struct {
 
 Capability provider MUST succeed request with a `"done"` status if provider is able to store requested [CAR] on user behalf right-away. In other words provider already has addressed [CAR] file.
 
+Capability provider MUST set `allocated` field to number of bytes it had no allocate for the [CAR], which MUST be either equal to `nb.size` of the invocation or `0` if no additional memory had to be allocated.
+
 ###### Store Add Upload
 
 Capability provider MUST succeed request with a `"upload"` status if it is able to allocate memory for the requested [CAR]. It MUST set `url` field to an HTTP PUT endpoint where addressed [CAR] can be uploaded.
@@ -202,6 +207,9 @@ Capability provider MUST succeed request with a `"upload"` status if it is able 
 Provider MUST set `headers` field to the set of HTTP headers that uploading agent MUST set on an HTTP PUT request.
 
 HTTP PUT endpoint set in `url` field MUST verify that uploaded bytes do correspond to the addressed [CAR] and specified `size`.
+
+Capability provider MUST set `allocated` field to number of bytes it had no allocate for the [CAR], which MUST be either equal to `nb.size` of the invocation or `0` if no additional memory had to be allocated.
+
 
 ###### Store Add Failure
 
--- a/w3-store.md
+++ b/w3-store.md
@@ -1,631 +1,870 @@
-# Storage and upload protocol
+# W3 Storage Protocol
 
 ![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
 
 ## Editors
 
 - [Yusef Napora](https://github.com/yusefnapora), [DAG House](https://dag.house)
+- [Irakli Gozalishvili](https://github.com/gozala)
 
 ## Authors
 
+- [Irakli Gozalishvili](https://github.com/gozala)
 - [Yusef Napora](https://github.com/yusefnapora), [DAG House](https://dag.house)
 
 ## Abstract
 
-This spec describes the data storage protocol used by web3.storage's "w3up" platform.
+In the W3 protocol user owned (name)space represent a data storage primitive that can be managed using W3 storage protocol defined here. Storage protocol allows space owners to manage state across compatible storage provider services using defined set of [UCAN] capabilities. Use of [UCAN] authorization system enables space access to be shared by delegating corresponding capabilities to desired audience.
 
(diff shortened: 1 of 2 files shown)


w3-store.md
```diff
@@ -1,631 +1,870 @@
...
 
-## Background
+# Introduction
 
-The w3up protocol accepts data in [CAR](https://ipld.io/specs/transport/car/) format, and clients are expected to encode data into CARs before invoking storage capabilities.
+The base storage layer of the space is [Content Address]ed [Content ARchive][CAR] files. In practice this means that user wishing to store files or directories of files needs produce an [IPLD] [DAG] in [UnixFS] format and then encoded into one or multiple [Content ARchive]s that can be stored using W3 storage protocol.
 
-The current w3up clients encode files and directories using [UnixFS](https://docs.ipfs.tech/concepts/file-systems/#unix-file-system-unixfs) and the [DAG-PB codec](https://ipld.io/specs/codecs/dag-pb/spec/), however, the storage protocol makes no assumptions about the format of the DAG (or DAGs) contained within a CAR, and any valid IPLD codec may be used.
+> Large DAGs get "sharded" across multiple [Content ARchive]s and stored individually to meet size limits of the storage provider.
```
vasco-santos, line 27:
```suggestion
> Large DAGs get "sharded" across multiple [Content ARchive]s and stored individually to meet potential size limits of the storage provider.
```

well, by our experience with old.web3.storage, I think we also concluded that more shards can be better for paralelization and more efficient retries on error. In addition, this also enables building a pause and resume upload pipeline which is great use case. I would suggest expanding this as not only a potential limitation of the storage provider

w3-store.md
```diff
@@ -1,631 +1,870 @@
...
-  origin?: string,
+Capability provider MUST issue `StoreGetSuccess` result for every content archive that has been added to the space.
 
-  /** ISO-8601 timestamp when CAR was added to the space */
-  insertedAt: string,
+```ipldsch
+type StoreGetSuccess {
+  link            &ContentArchive
+  size            Int
+  origin optional &ContentArchive
```
vasco-santos, line 286: should we comment that it is deprecated?

w3-store.md
```diff
@@ -1,631 +1,870 @@
...
+  StoreRemoveSuccess   "ok"
+  StoreRemoveFailure   "error"
+} representation keyed
+```
 
-### `store/list`
+###### Store Remove Success
 
-> Obtain a list of stored CARs
+Capability provider MUST issue `StoreRemoveSuccess` after it has removed the archive from the space.
```
vasco-santos, line 367:
```suggestion
Capability provider MUST issue `StoreRemoveSuccess` after it unlinked the archive from the space.
```
I think using unlined would be better, so that it does not get strictly understood as content won't be available anymore on the network

w3-store.md
```diff
@@ -1,631 +1,870 @@
...
+Authorized agent MAY invoke `upload/add` capability on the [space] subject (`with` field) to request that provider (`aud` field) include content identified by `nb.root` in the list of content entries for the space.
 
-An `upload/add` invocation requires the root CID of the DAG to register as an "upload," along with the CID of one or more CARs that contain the complete DAG. It is expected that these CARs have previously been stored in the space using [`store/add`](#storeadd); a service provider MAY return an error response if this is not the case.
+It is expected that CARs containing content are stored in the space using [`store/add`] capability. Provider MAY enforce this invariant by failing invocation or choose to succeed invocation but fail to serve the content when requested.
 
-#### Derivations
+> ⚠️ Behavior of calling `upload/add` with a same `root` and different `shards` is not specified by the protocol. w3up reference implementation allows such invocations and updates `shards` to union of all shards across invocations.
```
vasco-santos, line 505: maybe worth adding a note that there may exist limit of shards at implementation level?

w3-store.md
```diff
@@ -1,631 +1,870 @@
...
+```
 
-`upload/remove` can be invoked to remove the link between an uploaded data CID and the CARs containing the data.
+### Upload Remove
 
-Note that this will not remove the stored CARs; clients will need to use [`store/remove`](#storeremove) to remove the CARs once all uploads referencing those CARs have been removed.
+Authorized agent MAY invoke `upload/remove` capability to remove upload entry from the list in the specified space (`with` field).
 
-The `with` resource URI must be set to the DID of the space to remove the upload from.
+> ⚠️ Please note that removing upload entry SHOULD NOT remove [content archive][CAR]s containing contain.
```
vasco-santos, line 671: should this be a MUST instead?

vasco-santos (approved):
LGTM (small nits inline)

w3-store.md
```diff
@@ -1,631 +1,870 @@
-# Storage and upload protocol
+# W3 Storage Protocol
 
 ![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
```
vasco-santos, line 3: SHould we state that this is in deprecation stage?

w3-store.md
```diff
@@ -1,631 +1,870 @@
...
+  StoreAddDone      "done"
+  StoreAddPending   "upload"
+} representation inline {
+  discriminantKey "status"
+}
 
-On success, the response will include a `status` field, which may have one of the following values:
+type StoreAddDone struct {
+  with    SpaceDID
+  link    &ContentArchive
```
vasco-santos, line 178: we miss allocated property (this is used for billing to get to know how much to bill to the space). see:
- https://github.com/web3-storage/w3up/blob/main/packages/upload-api/src/store/add.js#L67
- https://github.com/web3-storage/w3up/pull/1213

w3-store.md
```diff
@@ -1,631 +1,870 @@
...
+}
+
+type StoreAddPending struct {
+  with      SpaceDID
+  link      &ContentArchive
```
vasco-santos, line 183: we miss allocated property (this is used for billing to get to know how much to bill to the space). see:
- https://github.com/web3-storage/w3up/blob/main/packages/upload-api/src/store/add.js#L67
- https://github.com/web3-storage/w3up/pull/1213

vasco-santos (changes requested):
actually this is required to be added

w3-store.md
Gozala: I don't think we should deprecate until we have alternative

Gozala, line 3:
```suggestion
![reliable](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square)
```

w3-store.md
Gozala, line 286:
```suggestion
  # deprecated
  origin optional &ContentArchive
```

w3-store.md
Gozala:
```suggestion
> ⚠️ Please note that removing upload entry MUST NOT remove [content archive][CAR]s containing contain.
```

vasco-santos (approved):

PR #115, opened 2024-03-19
feat: blob protocol draft

Gozala:
Based on #114. PR is written with UCAN 1.0 format and assuming https://github.com/web3-storage/RFC/pull/12 however in terms of immediate implementation I suggest that we instead

1. Use `blob/add` instead of `/space/content/add/blob`
2. Use `web3.storage/blob/*` in place of `/service/blob/` 

I suggest above because I think it would make more sense to transition to https://github.com/web3-storage/RFC/pull/12 once we have UCAN@1.0 implemented, because I suspect links to tasks vs invocations are going to be a pain to deal with otherwise. This will give us cleaner break.

In terms of implementing `/service/blob/accept` and specifically how does client signal that they've completed upload I suggest that we do whichever is easiest option from following two:

1. Make client sent second `blob/add` invocation after they've done upload so we can perform a lookup.
2. Add another temp capability with empty output either very specific like `blob/add/poll` or very general like `invocation/wake { task: CID } `.

### DIFF
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -167,3 +167,5 @@ SHA2-256
 noop
 Pre
 unlinked
+2MiB
+ok
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -152,3 +152,7 @@ fieldAccountPrincipal
 installationIssuerPrincipal
 iteratively
 base32
+AccountA
+AgentA
+IssuerA
+AudiencePrincipal
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -162,4 +162,7 @@ Uint8Array
 ed25519
 multihash
 IPNI
-ARchive
+aRchive
+SHA2-256
+noop
+Pre
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -160,3 +160,6 @@ dag-json-encoded
 base64url-multibase-encoded
 Uint8Array
 ed25519
+multihash
+IPNI
+ARchive
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -148,3 +148,6 @@ NameDescriptionPrincipalThe
 fieldIssuerPrincipal
 fieldAudiencePrincipal
 verifier
+fieldAccountPrincipal
+installationIssuerPrincipal
+iteratively
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -156,3 +156,7 @@ AccountA
 AgentA
 IssuerA
 AudiencePrincipal
+dag-json-encoded
+base64url-multibase-encoded
+Uint8Array
+ed25519
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -151,3 +151,4 @@ verifier
 fieldAccountPrincipal
 installationIssuerPrincipal
 iteratively
+base32
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -144,3 +144,7 @@ RFC2181
 RFC2822
 SMTP
 publicness
+NameDescriptionPrincipalThe
+fieldIssuerPrincipal
+fieldAudiencePrincipal
+verifier
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -166,3 +166,4 @@ aRchive
 SHA2-256
 noop
 Pre
+unlinked
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -33,7 +33,6 @@ There are several distinct roles that [principal]s may assume in this specificat
 | Issuer | A [principal] delegating capabilities to another [principal]. It is the signer of the [UCAN]. Specified in the `iss` field of a UCAN. |
 | Audience | Principal access is shared with. Specified in the `aud` field of a UCAN. |
 
-
 ### Space
 
 A namespace, often referred as a "space", is an owned resource that can be shared. It corresponds to a unique asymmetric cryptographic keypair and is identified by a [`did:key`] URI.
@@ -190,10 +189,9 @@ Blob can be published by authorizing read interface (e.g. IPFS gateway) by re-de
 
 > Note that same applies to publishing blob on [IPNI], new capability is not necessary, user simply needs to re-delegate `LocationClaim` to the DID representing [IPNI] publisher. [IPNI] publisher in turn may publish delegation to DID with publicly known private key allowing anyone to perform the reads.
 
-
 [store protocol]:./w3-store.md
 [CAR]:https://ipld.io/specs/transport/car/
 [multihash]:https://github.com/multiformats/multihash
 [space]:#space
 [referential transparency]:https://en.wikipedia.org/wiki/Referential_transparency
-[IPNI]:https://github.com/ipni/specs/blob/main/ipni.md
+[IPNI]:https://github.com/ipni/specs/blob/main/IPNI.md
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -77,7 +77,7 @@ Accepted -- claim --> Claim
 Added -- claim --> Accept
 ```
 
-**Iconography**
+#### Iconography
 
 - Icon on the left describes type of the node
 - Icon on the right describes issuer / audience pair if only one audience is the issuer
@@ -162,6 +162,7 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
             "content": { "/": { "bytes": "mEi...sfKg" } },
             "size": 2_097_152,
           },
+          exp: 1711122994101
           // This task is blocked on allocation
           _: { "await/ok": { "/": "bafy...alloc" } }
         }
@@ -413,7 +414,7 @@ Blob can be published by authorizing read interface (e.g. IPFS gateway) by deleg
 [location claim]:#location-claim
 [Add Blob]:#add-blob
 [Allocate Blob]:#allocate-blob
-[Accept Blob]:#receive-blob
+[Accept Blob]:#accept-blob
 [DID]:https://www.w3.org/TR/did-core/
 [Link]:https://ipld.io/docs/schemas/features/links/
 [range request]:https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -1,0 +1,199 @@
+# W3 Blob Protocol
+
+![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+- [Irakli Gozalishvili](https://github.com/gozala)
+
+## Authors
+
+- [Irakli Gozalishvili](https://github.com/gozala)
+
+## Abstract
+
+W3 blob protocol allows authorized agents to store arbitrary content blobs with a storage provider.
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+# Introduction
+
+W3 blob protocol provides core building block for storing content and sharing access to it through UCAN authorization system. It is successor to the [store protocol] which no longer requires use of [Content Archive][CAR]s even if in practice clients can continue to use it for storing shards of large DAGs.
+
+## Concepts
+
+### Roles
+
+There are several distinct roles that [principal]s may assume in this specification:
+
+| Name        | Description                                                                                                                                    |
+| ----------- | ---------------------------------------------------------------------------------------------------------------------------------------------- |
+| Principal | The general class of entities that interact with a UCAN. Identified by a DID that can be used in the `iss` or `aud` field of a UCAN. |
+| Agent       | A [Principal] identified by [`did:key`] identifier, representing a user in an application. |
+| Issuer | A [principal] delegating capabilities to another [principal]. It is the signer of the [UCAN]. Specified in the `iss` field of a UCAN. |
+| Audience | Principal access is shared with. Specified in the `aud` field of a UCAN. |
+
+
+### Space
+
+A namespace, often referred as a "space", is an owned resource that can be shared. It corresponds to a unique asymmetric cryptographic keypair and is identified by a [`did:key`] URI.
+
+### Blob
+
+Blob is a fixed size byte array addressed by the [multihash]. Usually blobs are used to represent set of IPLD blocks at different byte ranges.
+
+# Capabilities
+
+## Add Blob
+
+Authorized agent MAY invoke `/space/content/add/blob` capability on the [space] subject to store specific byte array.
+
+> Note that storing a blob does not imply advertising it on the network or making it publicly available.
+
+### Add Blob Capability
+
+#### Add Blob Capability Schema
+
+```ipldsch
+type struct AddBlob {
+  cmd   "/space/content/add/blob"
+  sub   SpaceDID
+  args  Blob
+}
+
+type struct Blob {
+  content   Multihash
+  size      Int
+}
+```
+
+#### Blob Content
+
+Blob `content` field MUST be a [multihash] digest of the blob payload bytes, uniquely identifying blob.
+
+#### Blob Size
+
+Blob `size` field MUST be set to the size of the blob in bytes.
+
+### Add Blob Receipt
+
+#### Add Blob Result
+
+```ipldsch
+type BlobAddResult union {
+  BlobAddSuccess    "ok"
+  BlobAddFailure    "error"
+} representation keyed
+
+type BlobAddSuccess union {
+  BlobAddAllocation     "allocated"
+  &UCANLocationClaim    "committed"
+} representation keyed
+
+type BlobAddAllocation struct {
+  content   Multihash
+  size      Int
+
+  url       URL
+  headers   HTTPHeaders
+}
+
+type UCANLocationClaim struct {
+  iss     DID
+  aud     DID
+  sub     SpaceDID
+  cmd     "assert/location"
+  args    LocationClaim
+  nonce   bytes
+  exp     Int
+}
+
+type LocationClaim struct {
+  content    Multihash
+  url        URL
+  range      ByteRange
+}
+
+type ByteRange struct {
+  offset    Int
+  length    Int
+} representation tuple
+```
+
+#### Add Blob Success
+
+##### Add Blob Allocated
+
+Capability provider MUST issue receipt with `BlobAddAllocation` success result if underlying space has enough capacity for the requested blob, but has not such blob locally.
+
+The `url` field MUST be set to the HTTP PUT endpoint where `content` matching specified [multihash] and `size` could be uploaded.
+
+The `headers` field MUST be set to the HTTP headers to be send along with `content` to the HTTP PUT endpoint.
+
+###### Add Blob Committed
+
+Capability provider MUST issue receipt with `LocationClaim` capability delegated to the [space] when requested blob is added to the space, which can happen if provider already has a matching blob locally.
+
+By delegating `LocationClaim` to the space provider makes a commitment to serve corresponding blob at byte range in from given URL to all authorized agents.
+
+> Not that since `aud` of the claim is the `space` DID, space can authorize other principals by re-delegating this claim.
+
+## Get Blob
+
+Authorized agent MAY invoke `/space/content/get/blob` capability on the [space] subject to query a state of the corresponding blob.
+
+### Get Blob Capability
+
+#### Get Blob Capability Schema
+
+```ipldsch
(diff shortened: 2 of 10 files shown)


w3-blob.md
```diff
@@ -0,0 +1,199 @@
...
+} representation keyed
+
+type BlobAddSuccess union {
+  BlobAddAllocation     "allocated"
+  &UCANLocationClaim    "committed"
+} representation keyed
+
+type BlobAddAllocation struct {
+  content   Multihash
+  size      Int
```
vasco-santos, line 95: I assume this is the size of the blob. we miss allocated property (this is used for billing to get to know how much to bill to the space). see:
- https://github.com/web3-storage/w3up/blob/main/packages/upload-api/src/store/add.js#L67
- https://github.com/web3-storage/w3up/pull/1213

vasco-santos: For extra content on how we use this. We used `allocated` for billing and it is important because it can either be:
- size of the bytes of new upload in the space
- size of the bytes of new upload already existing in other space (not uploaded, but should be accounted for space bill)
- 0 if already stored in space

w3-blob.md
```diff
@@ -0,0 +1,199 @@
...
+
+```ipldsch
+type BlobAddResult union {
+  BlobAddSuccess    "ok"
+  BlobAddFailure    "error"
+} representation keyed
+
+type BlobAddSuccess union {
+  BlobAddAllocation     "allocated"
+  &UCANLocationClaim    "committed"
```
vasco-santos, line 90: Is this for returning when we already have content? Otherwise it should be async right?

in other words, the `blob/close` kind of capability executed when we see that upload made it to write target

w3-blob.md
```diff
@@ -0,0 +1,199 @@
...
+
+Capability provider MUST issue receipt with `BlobAddAllocation` success result if underlying space has enough capacity for the requested blob, but has not such blob locally.
+
+The `url` field MUST be set to the HTTP PUT endpoint where `content` matching specified [multihash] and `size` could be uploaded.
+
+The `headers` field MUST be set to the HTTP headers to be send along with `content` to the HTTP PUT endpoint.
+
+###### Add Blob Committed
+
+Capability provider MUST issue receipt with `LocationClaim` capability delegated to the [space] when requested blob is added to the space, which can happen if provider already has a matching blob locally.
```
vasco-santos, line 135: I am confused here. This should only happen when bytes make its way to the bucket right? (based also on the diagram you previously shared)

w3-blob.md
```diff
@@ -0,0 +1,199 @@
...
+
+#### Get Blob Success
+
+Capability provider MUST return the state of the requested blob. If memory for the blob was allocated but content has not been yet uploaded result MUST be of `BlobAddAllocation` case. If blob has been allocated and uploaded result MUST of `UCANLocationClaim` case.
+
+> Please note that implementation MAY check local state on invocation and update from `BlobAddAllocation` to `UCANLocationClaim`, however this SHOULD be done in referentially transparent way. In other words no side effects should be exhibited, instead function should lazily compute and cache.
+
+## Publish Blob
+
+Blob can be published by authorizing read interface (e.g. IPFS gateway) by re-delegating `LocationClaim` for a corresponding blob to it.
```
vasco-santos, line 189: I think we need more details here. This is supposed to be for the location of the multihash of the blob, or every single block within the blob (which would mainly happen in IPNI use case)

vasco-santos (changes requested):
Thanks for creating this ❤️  we are generally quite close on it

hannahhoward (commented):
Is the intent for blob/* to eventually replace store/*? 

Do we have the work to implement this tracked anywhere? (seems like it's not trivial amount of work -- should probably ship with spec)

hannahhoward (changes requested):
Think we need agreement at minimum on prioritizing `blob/*` work in client and server before we merge this spec. Otherwise it's a set of actions that aren't implemented

vasco-santos:
@hannahhoward yes! This is our way forward with write to anywhere to avoid the migration pains. But we need to include in the write to anywhere ticket

w3-blob.md
Gozala: Yeah this was basically what we have right now in the store/* API, although I do think you might be raising a good point here perhaps we need same result and this should be linked through an effect.

It might help to go over several options so we can compare tradeoffs and choose a right compromise.

w3-blob.md
Gozala: Aren't first two effectively the same ? I would expect that we will not have to see `0` because in that case you'll get `commited` variant as opposed to `allocated`. Am I missing something ?

w3-blob.md
Gozala: I think this is poorly worded, I'll fix it but what I meant to say that if provider already has such blob they can issue `"committed"` case of the ok result as opposed to `"allocated"`.

You are right that if provider does not already have a blob we are not going to issue `LocationClaim` until we know we have it.

w3-blob.md
Gozala: No it is the whole blob, equivalent of accessing car from gateway by CAR cid, except in this case it would be raw CID. Block level index can be then accomplished through inclusions claims that effectively say that this multihash is a slice of that other multihash at the given byte offset and this number of bytes

Gozala:
> Is the intent for blob/* to eventually replace store/*?
> 
> Do we have the work to implement this tracked anywhere?

I do think we want to some day, but there is no urgency to do so. We do want to introduce new blob capabilities here as they do not imply bucket events and we want to migrate clients to that.

>  (seems like it's not trivial amount of work -- should probably ship with spec)

Generally we have been aligning on spec first and only then going about implementing it, sometimes that imply changes to spec but alignment on how seemed to be a good way to avoid changing code back and forth.

Gozala (commented):
comment

w3-blob.md
Gozala: TODO: How do we deal with /add → /remove → /add if we say we issue the same receipt.

Gozala:
```js
{
  ran: { "/": "bafy..add" },
  out: { ok: { allocated: 123 } },
  next: [
    { cmd: "service/blob/wait", with: "did:web:web3.storage", args: { content: 'hash', size: 123 }  },
    { cmd: "assert/location", with: "did:web:web3.storage", args: { url: "https://...", headers: {}, _: { "await/ok": {"/": "bafy...blobawait" } } } } 
  ]

}
```

Gozala:
```js
{
  iss: "did:web:web3.storage",
 aud: did:key:alice
  att: [{
     can: assert/location
      with: "did:web:web3.storage",
      nb: {
          content: {},
          url: ...
          headers: ...
          range: [0, 123]
      }
  }]
}
```

w3-blob.md
```diff
@@ -0,0 +1,199 @@
...
+
+Capability provider MUST return the state of the requested blob. If memory for the blob was allocated but content has not been yet uploaded result MUST be of `BlobAddAllocation` case. If blob has been allocated and uploaded result MUST of `UCANLocationClaim` case.
+
+> Please note that implementation MAY check local state on invocation and update from `BlobAddAllocation` to `UCANLocationClaim`, however this SHOULD be done in way that preserves [referential transparency]. In other words no side effects should be observable, instead function should lazily compute and cache.
+
+## Publish Blob
+
+Blob can be published by authorizing read interface (e.g. IPFS gateway) by re-delegating `LocationClaim` for a corresponding blob to it.
+
+> Note that same applies to publishing blob on [IPNI], new capability is not necessary, user simply needs to re-delegate `LocationClaim` to the DID representing [IPNI] publisher. [IPNI] publisher in turn may publish delegation to DID with publicly known private key allowing anyone to perform the reads.
```
hannahhoward, line 191: So once something is published to IPNI, the provider (i.e. W3S) needs to keep updating it without having the user redelegate every time. It's not clear if this flow supports that.

w3-blob.md
hannahhoward: What if the Blob gets moved around by w3s? Does it need to be redelegated?

w3-blob.md
Gozala: did:web:web3.storage → (service key, space did) → did:key:alice

w3-blob.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Add Blob Capability
+
+#### Add Blob Capability Schema
+
+```ipldsch
+type struct AddBlob {
+  cmd   "/space/content/add/blob"
+  sub   SpaceDID
+  args  Blob
```
Gozala, line 61: We should allow user specify an expiry here so we can reflect it on our end.

hannahhoward (changes requested):
For now, I think we should simply make the audience of the location claim public. The current flow implies a redelegation of authority every time the location changes.

I'm concerned we're thinking we can get private data for free here without thinking through the implications. Private retrieval is a huge topic -- I think we should design for it intentionally, rather than throw in features that may create lots of headaches on a theory it will be sufficient for private data. I think privacy should be set at the ingress point (i.e. by the space owner on the blob itself, absent the location, either during blob/add or through some seperate update apit), not on the location claim itself which is temporary.

Either way, I think we should move it out of our current design, because this requires more intention.

hannahhoward:
Realizing removing the redelegation requires invoking `IPNI/offer` and/or `store/publish` as seperate capabilities. Again, this is I think the right approach.

If a location is managed by the provider, it must be responsibile for keeping location claims, wherever they reside, updated.

Gozala:
> If a location is managed by the provider, it must be responsibile for keeping location claims, wherever they reside, updated.

What we settled on in the https://github.com/web3-storage/RFC/pull/13 that our issued location claims will have HTTP URLs with hints of where content is. Those URLs will simply route / redirect to the location content is in the system.

This enables us to make long term location commitments while retaining ability to change actual site of the content over time.

Does this address the concern you're raising ?

Gozala:
> For now, I think we should simply make the audience of the location claim public. The current flow implies a redelegation of authority every time the location changes.

I think this is inaccurate, as mentioned in prior comment per https://github.com/web3-storage/RFC/pull/13 we intend to make issued location changes agnostic of content site changes.


> I'm concerned we're thinking we can get private data for free here without thinking through the implications. Private retrieval is a huge topic -- I think we should design for it intentionally, rather than throw in features that may create lots of headaches on a theory it will be sufficient for private data. I think privacy should be set at the ingress point (i.e. by the space owner on the blob itself, absent the location, either during blob/add or through some seperate update apit), not on the location claim itself which is temporary.

I think there is a misunderstanding here. I also want to assure you that I have been thinking about private retrievals for more than a year now it's just there were always higher priorities at play.

I'd like to step back from privacy here for moment. What is proposed here is not to assume that adding a blob implies you want it to be indexed, advertised and served across all possible channels. Instead we can issue a commitment to the space itself that we will serve the blob from the given URL through the commitment validity window. This commitment can be published by user if they want to make that data available publicly. This also creates an opportunity to create a system where our commitments can be verified and we can be held accountable if we fail to uphold them.

In other words it gives user a choice to do what they want with our commitment to serve the blob.

w3-blob.md
Gozala: No it does not per https://github.com/web3-storage/RFC/pull/13 we intend to use locations that are agnostic of actual blob site, they can HTTP redirect as necessary. Also more intelligent clients could recognize hints in the URL to go directly to a site without going through a redirect. The best thing is we always have a fallback and ability to changes things without having to distribute updates.

w3-blob.md
Gozala: Please bare with me as this spec is by no means done and things are not very clear right now. To answer your question however, yes publisher can keep publishing advertisements as long as delegation is valid. If it had been time limited that is in fact was a user intention.

hannahhoward:
Thinking back to https://github.com/web3-storage/RFC/pull/13 I now see this is a bit of a misnomer as a solution.

Yes, it's a public URL that will continue to work, if you go through w3s.link. 

It still has an expiration and the data still could move. Presumably if I have `https://w3s.link/ipfs/bafy...BLOBCID?origin=r2://region/bucketName/key` and then it gets moved, presumably the service would want to  publish at minimum a second claim with a better hint -- i.e. `https://w3s.link/ipfs/bafy...BLOBCID?origin=r2://region2/bucketName2/key2` -- how does the client know to re-delegate this new claim in order to publish it? Or if the claim expires, and the publisher wants the new claim (potentially still at the same location) re-published, are they in charge of monitoring that?

Also, unless I misunderstand, we are trying to get to a public mechanism whereby a client can find the location of the car and its index, and do range reads. How would it do so for this case if the hints become out of date or the claim expires? Would it talk directly to the w3s content claims service? How would it know to do so (assuming it wasn't a web3.storage exclusive retrieval client)? Or be forced to fetch from w3s.link instead?

Ultimately, I think we're dealing with two different chains of authority we're trying to smash into one here:
1. Authority to manage the visibility / discoverability of a piece of data
2. Authority to move that data around and update its location

I think we should probably take a step back and sync on all this.  I feel that these are valid concerns, not just nitpicking, especially as we look forward to the architecture we're building.

hannahhoward (approved):
Removing objections after synchronous discussion. Will document reasoning shortly.

hannahhoward:
Documentation about why I removed my objections:

- A public location is in fact a commitment to make data available at a given URL for the length of the claim. Therefore, it makes sense that it's ok to let the user choose to then share the claim with the wider audience
- https://github.com/web3-storage/RFC/pull/13 is intended to enable this - in the sense that it enables the provider to make a claim that data is available at a given URL, and maintain the URL in a context where it moves around internally
- the same PR is not likely a complete solution for the point where we have a network of providers. But that's ok -- we can explore that when we get there, either by having those providers publish claims themselves, or by using w3s as a router in claims. There is probably more design work to do here.

w3-blob.md
```diff
@@ -0,0 +1,419 @@
...
+  sub: SpaceDID
+  args: {
+    blob: Blob
+  }
+}
+
+type Blob = {
+  content: Multihash
+  size:    int
+}
```
vasco-santos, line 190: We should allow user specify an expiry here so we can reflect it on our end with `accept`.

w3-blob.md
```diff
@@ -0,0 +1,419 @@
...
+      // if matches blob multihash and size.
+      { // "/": "bafy...accept",
+        "cmd": "/service/blob/accept",
+        "sub": "did:web:web3.storage",
+        "args": {
+          "space": "did:key:zAlice",
+          "blob": {
+            // multihash of the blob as byte array
+            "content": { "/": { "bytes": "mEi...sfKg" } },
+            "size": 2_097_152,
```
vasco-santos, line 163: expiration here?

w3-blob.md
```diff
@@ -0,0 +1,419 @@
...
+
+The `url` of the `BlobAddress` MUST be an HTTP(S) location that can receive blob content via HTTP `PUT` request, as long as HTTP headers from `headers` dictionary are set on the request.
+
+It is RECOMMENDED that issued `BlobAddress` only accept `PUT` payload that matches requested `blob`, both content [multihash] and size.
+
+> ℹ️ If enforcing above recommendation is not possible implementation MUST enforce in [Accept Blob] invocation.
+
+### Allocation Effects
+
+Allocation MUST have no effects.
```
vasco-santos, line 325: I was thinking this would actually point to Accept as well, given next step is that after allocation. Why not?

vasco-santos (approved):
There are a few details in comment, as well as remove/list to add here. But I am happy to cover them in follow up :)

w3-blob.md
```diff
@@ -0,0 +1,419 @@
...
+}
+```
+
+#### Allocation Space
+
+The `args.space` field MUST be set to the [DID] of the user space where allocation takes place.
+
+#### Allocation Blob
+
+The `args.blob` field MUST be set to the `Blob` the space is allocated for.
```
vasco-santos, line 276: this is content right? Should we keep same name?

w3-blob.md
```diff
@@ -0,0 +1,419 @@
...
+Add --o Added
+Added -.-> Allocate
+Added -.-> Accept
+Allocate --> Accept
+Accept --o Accepted
+Accepted -- claim --> Claim
+Added -- claim --> Accept
+```
+
+**Iconography**
```
Gozala, line 80:
```suggestion
#### Iconography
```

.github/workflows/words-to-ignore.txt
```diff
@@ -160,3 +160,10 @@ dag-json-encoded
 base64url-multibase-encoded
 Uint8Array
 ed25519
+multihash
+IPNI
+aRchive
+SHA2-256
+noop
+Pre
+unlinked
```
Gozala, line 169:
```suggestion
unlinked
2MiB
```

w3-blob.md
```diff
@@ -0,0 +1,419 @@
...
+      { // "/": "bafy...accept",
+        "cmd": "/service/blob/accept",
+        "sub": "did:web:web3.storage",
+        "args": {
+          "space": "did:key:zAlice",
+          "blob": {
+            // multihash of the blob as byte array
+            "content": { "/": { "bytes": "mEi...sfKg" } },
+            "size": 2_097_152,
+          },
```
Gozala, line 164:
```suggestion
          },
          exp: 1711122994101
```

w3-blob.md
Gozala: I think you may have misunderstood my intention behind `exp` in `accept`, there it meant to capture pre-signed url expiry. While here I would expect it to imply that blob should be GC-ed at given time.

Am I interpreting your intentions correctly here ? If so I'd say lets not add this until we actually are able to support GC-ing blobs, when we do we can introduce this field.

w3-blob.md
```diff
@@ -0,0 +1,419 @@
...
+[store protocol]:./w3-store.md
+[CAR]:https://ipld.io/specs/transport/car/
+[multihash]:https://github.com/multiformats/multihash
+[space]:#space
+[IPNI]:https://github.com/ipni/specs/blob/main/IPNI.md
+[await/ok]:https://github.com/ucan-wg/invocation?tab=readme-ov-file#await
+[location claim]:#location-claim
+[Add Blob]:#add-blob
+[Allocate Blob]:#allocate-blob
+[Accept Blob]:#receive-blob
```
Gozala, line 416:
```suggestion
[Accept Blob]:#accept-blob
```

w3-blob.md
Gozala: Perhaps 🤔, but here here is my sentiment

Can I allocate space without waiting for it to be received ? If so, this is a function to do and it should not have a continuation there. If no, I think I need bit more convincing because here are my counter arguments

1. Overall I think you could roughly translate blob/add as following JS code

   ```js
   const blobAdd = async (task) => {
      const { sub, args: { blob, exp } } = task
      // check that we have billing setup for the space
      const { ok: subscription, error: cantbill } = await service.getSubscription(sub)
      if (cantbill) {
         return { error: cantbill }
      }

      // allocate memory for the blob
      const { ok: memory, error: allocError } = await service.allocate(sub, { exp, blob, cause: task.link() })
      if (allocError) {
        return { error: allocError }
      }
      
      // add usage to the billing
      if (memory.size > 0) {
         await subscription.bill.add({ storage: size, cause: task.link() })
      }
      
       // if we have memory address we need to await for the content
      if (memory.address) {
         return { ok: await accept({ blob, exp  }) }
      } else {
         return { ok: findReceiptFor({ accept, blob, exp }) }
      }
    }

    const accept = async ({ blob, exp }) => {
       const { ok: content, error: timeout } = await waitForPUT({ blob })
        if (timeout) {
           return { error: timeout }
        }
          
        // validate content matches the blob
        const { error } = await vaildate(content, blob)
        if (error) {
            return { error }
        }

        return { ok: createLocationClaim(blob) })
    }
    ``` 

    And through above lens, `addBlob` is a task that coordinates subtasks, but subtasks `allocate` and `accept` are independent of each other.

2. If we made `accept` effect of the `allocate`, then we probably should not make it effect of the `blobAdd` also. However making it effect of the `blobAdd` is better because we can then draw the whole execution graph from the receipt without having to follow each effect.


For what it's worth these observations and opinions are new to me also and  I have call with Fission on Monday to try and get their input on this as they spend a lot more time doing and implementing this.

w3-blob.md
Gozala: It references the `Blob` type defined earlier which is `{ content, size }` pair

w3-blob.md
Gozala: Added below

w3-blob.md
```diff
@@ -0,0 +1,420 @@
...
+        "cmd": "/service/blob/accept",
+        "sub": "did:web:web3.storage",
+        "args": {
+          "space": "did:key:zAlice",
+          "blob": {
+            // multihash of the blob as byte array
+            "content": { "/": { "bytes": "mEi...sfKg" } },
+            "size": 2_097_152,
+          },
+          exp: 1711122994101
```
Gozala, line 165:
```suggestion
          exp: 1711122994101,
```

w3-blob.md
vasco-santos: oh I see, ok that is great then 👍🏼

w3-blob.md
vasco-santos: yes, I was considering this to be the expiration we talked about for expiration of claim https://github.com/web3-storage/specs/pull/115#discussion_r1530787111

but than can come as follow up

w3-blob.md
vasco-santos: thanks for the details, I find this a good reasoning. My suggestion was more about having a receipt for a `blob/allocate` and not really linking it to anything. But we have as part of the task arguments the `blob/add` who caused this, so I think that also mitigates what I was looking for

vasco-santos (approved):
LGTM

Eianaftyio:
.diff

PR #116, opened 2024-04-02
feat: introduce http/put capability

Gozala:
Amend spec per agreements last Thursday, few notes:

1. I end up describing `ucan/conclude` capability as opposed to `ucan/receipt`, although we could go with receipt route if it is more preferred. Either one will likely differ from where spec will end up, but current version is likely to be closer than version with `ucan/receipt`.
2. `await/ok` in the `out` field may change slightly as we're discussing doing something like `{ 'ucan/await': ['.ok.claim', { "/": ... }] }` so you could drill down deeper, however there is no agreement in UCAN WG on exact schema yet.

### DIFF
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -158,7 +158,7 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
       // 2. System requests user agent (or anyone really) to upload the content
       // corresponding to the blob
       // via HTTP PUT to given location.
-      { // "/": "bafy...upload",
+      { // "/": "bafy...put",
         "cmd": "/http/put",
         "sub": "did:key:zMh...der", // <-- Ed299.. derived key from content multihash
         "args": {
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -164,7 +164,7 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
         "args": {
           "content": { "/": { "bytes": "mEi...sfKg" } },
           "address": { "await/ok": { "/": "bafy...alloc" } },
-          _allocate: { "await/ok": { "/": "bafy...alloc" } }
+          "_allocate": { "await/ok": { "/": "bafy...alloc" } }
         },
         "meta": {
           // archive of the principal keys
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -368,8 +368,9 @@ type BlobPut = {
   cmd: "/http/put"
   sub: DID
   args: {
-    content: Multihash
-    address: BlobAddress
+    url: URL
+    headers: Headers
+    body: Blob
   }
   meta: {
     keys: {[key: DID]: bytes}
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -163,7 +163,8 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
         "sub": "did:key:zMh...der", // <-- Ed299.. derived key from content multihash
         "args": {
           "content": { "/": { "bytes": "mEi...sfKg" } },
-          "address": { "await/ok": { "/": "bafy...alloc" } }         
+          "address": { "await/ok": { "/": "bafy...alloc" } },
+          _allocate: { "await/ok": { "/": "bafy...alloc" } }
         },
         "meta": {
           // archive of the principal keys
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -61,6 +61,8 @@ Added("🧾 { ok: { claim 🚦 } }")
 
 Allocate("⏭️ /service/blob/allocate 🤖")
 
+Put("⏭️ /http/put 🤖🔑")
+
 Accept("⏭️ /service/blob/accept 🤖")
 
 Accepted("🧾 { ok: { claim 🎫 } }")
@@ -71,7 +73,10 @@ Claim("🎫 /assert/location 🤖👩‍💻")
 Add --o Added
 Added -.-> Allocate
 Added -.-> Accept
+Added -.-> Put
 Allocate --> Accept
+
+Put --> Accept
 Accept --o Accepted
 Accepted -- claim --> Claim
 Added -- claim --> Accept
@@ -150,8 +155,26 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
           }
         }
       },
-      // 2. System will attempt to accept received content
-      // if matches blob multihash and size.
+      // 2. System requests user agent (or anyone really) to upload the content
+      // corresponding to the blob
+      // via HTTP PUT to given location.
+      { // "/": "bafy...upload",
+        "cmd": "/http/put",
+        "sub": "did:key:zMh...der", // <-- Ed299.. derived key from content multihash
+        "args": {
+          "content": { "/": { "bytes": "mEi...sfKg" } },
+          "url": "https://r2.cloudflare.com/ipfs/bafy...",
+          "headers": {}
+        },
+        "meta": {
+          // archive of the principal keys
+          "keys": {
+            "did:key:zMh...der": { "/": "mEi...sfKg" } 
+          }
+        }
+      },
+      // 3. System will attempt to accept uploaded content that matches blob
+      // multihash and size.
       { // "/": "bafy...accept",
         "cmd": "/service/blob/accept",
         "sub": "did:web:web3.storage",
@@ -164,7 +187,8 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
           },
           exp: 1711122994101,
           // This task is blocked on allocation
-          _: { "await/ok": { "/": "bafy...alloc" } }
+          _allocate: { "await/ok": { "/": "bafy...alloc" } },
+          _put: { "await/ok", { "/": "bafy...put" } }
         }
       }
     ]
@@ -213,6 +237,7 @@ type AddBlobReceipt = {
   out: Result<AddBlobOk, AddBlobError>
   next: [
     AllocateBlob,
+    PutBlob,
     AcceptBlob,
   ]
 }
@@ -246,6 +271,7 @@ Task linked from the `claim` of the success value MUST be present in the receipt
 Successful invocation MUST start a workflow consisting of following tasks, that MUST be set in receipt effects (`next` field) in the following order.
 
 1. [Allocate Blob]
+1. [Put Blob]
 1. [Accept Blob]
 
 ## Allocate Blob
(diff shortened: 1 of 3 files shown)


w3-blob.md
```diff
@@ -150,8 +155,26 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
...
+      // 2. System requests user agent (or anyone really) to upload the content
+      // corresponding to the blob
+      // via HTTP PUT to given location.
+      { // "/": "bafy...upload",
+        "cmd": "/http/put",
+        "sub": "did:key:zMh...der", // <-- Ed299.. derived key from content multihash
+        "args": {
+          "content": { "/": { "bytes": "mEi...sfKg" } },
+          "url": "https://r2.cloudflare.com/ipfs/bafy...",
+          "headers": {}
```
vasco-santos, line 167: should we wrap this with an address key to be consistent with Allocate receipt?

w3-blob.md
```diff
@@ -325,11 +351,76 @@ It is RECOMMENDED that issued `BlobAddress` only accept `PUT` payload that match
...
+#### Put Blob Capability Schema
+
+> ℹ️ In the UCAN 0.9 `meta` is unknown as `fct` field
+
+```ts
+type BlobPut = {
+  cmd: "/http/put"
+  sub: DID
+  args: {
+    address: BlobAddress
```
vasco-santos, line 371: Looks like content is missing here?

vasco-santos: or at least, this is not in sync with above. I would like to see content here anyway, as we should be able to query for the content rather than needing to know how a write target encodes it in the address (which actually may even not happen)

vasco-santos (commented):
LGTM, there are small details to wrap up

w3-blob.md
```diff
@@ -375,13 +466,36 @@ Receipt MUST not have any effects.
...
+```js
+{
+  "iss": "did:web:web3.storage",
+  "aud": "did:key:zAlice",
+
+  "cmd": "/assert/location",
+  "sub": "did:web:web3.storage",
+  "pol": [
+    // multihash must match be for the blob uploaded
+    ["==", ".content", { "/": { "bytes": "mEi...sfKg" } }],
```
vasco-santos, line 480: I actually now realize that content claims won't work with content?

w3-ucan.md
```diff
@@ -203,6 +203,70 @@ It is RECOMMENDED to treat revocations permanent. Even though enclosing [UCAN] c
...
+type Conclusion struct {
+  ran         &Invocation
+  out         Result
+  next        [&Task]
+  meta        Meta
+  time        Int -- Unix timestamp at which the Receipt was issued
+}
+
+type Result union {
+  Any               "ok"
```
vasco-santos, line 235: This should contain a block CID with encoded receipt attached right? Probably it can be already specified as it should be the same for everything?

w3-blob.md
Gozala:
```suggestion
    content: Multihash
    address: BlobAddress
```

w3-blob.md
Gozala: Why not ?

w3-blob.md
```diff
@@ -413,8 +521,13 @@ Blob can be published by authorizing read interface (e.g. IPFS gateway) by deleg
...
+[Put Blob]:#put-blob
+[put blob receipt]:#put-blob-receipt
 [Allocate Blob]:#allocate-blob
 [Accept Blob]:#accept-blob
 [DID]:https://www.w3.org/TR/did-core/
 [Link]:https://ipld.io/docs/schemas/features/links/
 [range request]:https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests
+[`did:key`]:https://w3c-ccg.github.io/did-method-key/
+[Ed25519]:https://en.wikipedia.org/wiki/EdDSA#Ed25519
+[UCAN Conclusion]:#./w3-ucan.md#conclusion
```
Gozala, line 533:
```suggestion
[UCAN Conclusion]:./w3-ucan.md#conclusion
```

w3-blob.md
vasco-santos: Everything is supposed to be a link https://github.com/web3-storage/content-claims/blob/main/packages/core/src/capability/assert.js#L15 
we need to raise this with rest of the team, probably a good opportunity to already rename to commitment?

w3-blob.md
```diff
@@ -150,8 +155,28 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
...
+      // corresponding to the blob
+      // via HTTP PUT to given location.
+      { // "/": "bafy...upload",
+        "cmd": "/http/put",
+        "sub": "did:key:zMh...der", // <-- Ed299.. derived key from content multihash
+        "args": {
+          "content": { "/": { "bytes": "mEi...sfKg" } },
+          "address": {        
+            "url": "https://r2.cloudflare.com/ipfs/bafy...",
+            "headers": {}
```
vasco-santos, line 168: I now realize that this is not great :/ We can iterate after, but in the world of write to anywhere where in allocate we could look async for write targets to write, we now need to know these right away. I wonder what would be the downside of just passing  the content, as service will always need to validate it

w3-blob.md
```diff
@@ -150,8 +155,28 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
           }
         }
       },
-      // 2. System will attempt to accept received content
-      // if matches blob multihash and size.
+      // 2. System requests user agent (or anyone really) to upload the content
+      // corresponding to the blob
+      // via HTTP PUT to given location.
+      { // "/": "bafy...upload",
```
vasco-santos, line 161:
```suggestion
      { // "/": "bafy...put",
```

w3-blob.md
```diff
@@ -150,8 +155,28 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
...
+            "headers": {}
+          }
+        },
+        "meta": {
+          // archive of the principal keys
+          "keys": {
+            "did:key:zMh...der": { "/": "mEi...sfKg" } 
+          }
+        }
+      },
```
vasco-santos, line 177: while I understand and like the idea of using this, I am actually now failing to understand how this pans out in practise. If we provide effect with CID for this task on the receipt of `blob/add`, how can client receiving this receipt know the key to invoke `http/put`? 

So, I am only seeing that this should go on the receipt of `blob/add` as well, or client would need to find this task content stored somewhere (we currently only store task when it is received to be executed...)

w3-ucan.md
```diff
@@ -203,6 +203,70 @@ It is RECOMMENDED to treat revocations permanent. Even though enclosing [UCAN] c
...
+
+type Conclude struct {
+  with          Authority
+  nb            Conclusion
+}
+
+type Conclusion struct {
+  ran         &Invocation
+  out         Result
+  next        [&Task]
```
vasco-santos, line 229: we do not need this for this use case, but do we need to specify join and fork?

w3-blob.md
Gozala: I think there are several options we could consider

1. Make that field a type union so it it could be either CID or multihash.
1. Encode multihash as CID with raw codec
1. Use different cmd name
1. Upgrade path

We could also just use a different field name and deprecate old one. For what it's worth linked capability uses `location` field as opposed to `url` and that could be used as signal to differentiate between two.

I have no strong feeling about which option to go with, I just don't want to tie us to the decisions we made in the past

Gozala:
```js
class ContentLocationClaim {
   constructor(model) {
        this.model = model
   }
   get content() {
      if (this.model.content instanceof Uint8Array) {
          return Link.create(RAW, Digest.decode(htis.model.content))
      } else {
          return this.model.content
      }
   }
   // ....
}

File: w3-blob.md [hunk 10]
```diff
@@ -150,8 +155,28 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
...
+      // via HTTP PUT to given location.
+      { // "/": "bafy...upload",
+        "cmd": "/http/put",
+        "sub": "did:key:zMh...der", // <-- Ed299.. derived key from content multihash
+        "args": {
+          "content": { "/": { "bytes": "mEi...sfKg" } },
+          "address": {        
+            "url": "https://r2.cloudflare.com/ipfs/bafy...",
+            "headers": {}
+          }
```
Gozala, line 169:
```suggestion
          "address": { "await/ok": { "/": "bafy...alloc" } }         
```

w3-blob.md
```diff
@@ -325,11 +350,77 @@ It is RECOMMENDED that issued `BlobAddress` only accept `PUT` payload that match
...
+#### Put Blob Capability Schema
+
+> ℹ️ In the UCAN 0.9 `meta` is unknown as `fct` field
+
+```ts
+type BlobPut = {
+  cmd: "/http/put"
+  sub: DID
+  args: {
+    content: Multihash
```
vasco-santos, line 370: can we make this a Blob instead? when we receive a `http/put` having the `blob` allows us to invoke `blob/accept` right away. Otherwise, we need to look at state to find it out

w3-blob.md
Gozala: We agreed to keep data model as defined here and decide how to encode this store separately

w3-blob.md
vasco-santos: yes, body or Blob is good for me

w3-blob.md
```diff
@@ -150,8 +155,25 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
...
-      // if matches blob multihash and size.
+      // 2. System requests user agent (or anyone really) to upload the content
+      // corresponding to the blob
+      // via HTTP PUT to given location.
+      { // "/": "bafy...put",
+        "cmd": "/http/put",
+        "sub": "did:key:zMh...der", // <-- Ed299.. derived key from content multihash
+        "args": {
+          "content": { "/": { "bytes": "mEi...sfKg" } },
+          "address": { "await/ok": { "/": "bafy...alloc" } }
```
Gozala, line 166: Add dependency on allocation

```suggestion
          "address": { "await/ok": { "/": "bafy...alloc" } },
          _allocate: { "await/ok": { "/": "bafy...alloc" } }
```

w3-blob.md
```diff
@@ -150,8 +155,26 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
...
+      // 2. System requests user agent (or anyone really) to upload the content
+      // corresponding to the blob
+      // via HTTP PUT to given location.
+      { // "/": "bafy...put",
+        "cmd": "/http/put",
+        "sub": "did:key:zMh...der", // <-- Ed299.. derived key from content multihash
+        "args": {
+          "content": { "/": { "bytes": "mEi...sfKg" } },
+          "address": { "await/ok": { "/": "bafy...alloc" } },
+          _allocate: { "await/ok": { "/": "bafy...alloc" } }
```
Gozala, line 167:
```suggestion
          "_allocate": { "await/ok": { "/": "bafy...alloc" } }
```

w3-blob.md
```diff
@@ -325,11 +351,77 @@ It is RECOMMENDED that issued `BlobAddress` only accept `PUT` payload that match
...
+
+> ℹ️ In the UCAN 0.9 `meta` is unknown as `fct` field
+
+```ts
+type BlobPut = {
+  cmd: "/http/put"
+  sub: DID
+  args: {
+    content: Multihash
+    address: BlobAddress
```
Gozala, line 372:
```suggestion
    url: URL
    headers: Headers
    body: Blob
```

PR #117, opened 2024-04-08
fix: update http put args description

vasco-santos:
We updated this but did not change it all

### DIFF
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -2,11 +2,15 @@
 
 ![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
 
+## Editors
+
 - [Irakli Gozalishvili](https://github.com/gozala)
+- [Vasco Santos](https://github.com/vasco-santos)
 
 ## Authors
 
 - [Irakli Gozalishvili](https://github.com/gozala)
+- [Vasco Santos](https://github.com/vasco-santos)
 
 ## Abstract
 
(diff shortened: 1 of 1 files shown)


w3-blob.md
```diff
@@ -162,7 +166,13 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
...
         "args": {
-          "content": { "/": { "bytes": "mEi...sfKg" } },
+          "body": {
+            // multihash of the blob as byte array
+            "content": { "/": { "bytes": "mEi...sfKg" } },
+            // size of the blob in bytes
+            "size": 2_097_152,
+          }
+          // TODO: this should now be updated?
           "address": { "await/ok": { "/": "bafy...alloc" } },
```
vasco-santos, line 176: @Gozala this ended up changing below to use `url` and `headers` https://github.com/web3-storage/specs/blob/main/w3-blob.md#put-blob-capability-schema

I honestly liked more the previous alignment here with address, but happy to change this if you have strong preference to use `url` and `header` top level

vasco-santos: I assume if so, we would like to have something along the lines of 

`'ucan/await': ['.out.ok.claim', acceptfx.link()],`

Gozala: I end up updating it here https://github.com/web3-storage/specs/pull/118/files last week but forgot to submit a PR for the review. I do not feel very strongly about it, however I find `{ url, headers, body }` to be more obvious choice for `http/put` than `{ address, body }`.

Gozala:
I think this is superseded by #118 but let me know if you think it should not.

PR #118, opened 2024-04-08
chore: switch to more coherent naming

Gozala:
[Preview](https://github.com/web3-storage/specs/blob/fix/align-blob-with-impl/w3-blob.md)

### DIFF
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -123,7 +123,7 @@ Shown Invocation example illustrates Alice requesting to add 2MiB blob to her sp
   "iss": "did:key:zAlice",
   "aud": "did:web:web3.storage",
   "args": {
-    "site": {
+    "blob": {
       // multihash of the blob as byte array
       "digest": { "/": { "bytes": "mEi...sfKg" } },
       // size of the blob in bytes
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -51,48 +51,66 @@ Authorized agent MAY invoke `/space/content/add/blob` capability on the [space] 
 
 ### Add Blob Diagram
 
-Following diagram illustrates execution workflow.
+Following diagram illustrates execution flow. Alice invokes `/space/content/add/blob` command which produces receipt with three effects (`allocate`, `put`, `accept`) and awaited `claim`. Effects have dependencies and therefor predict execution flow (from left to right). The output of the main task awaits on the result of the result of the last effect.
 
 ```mermaid
 flowchart TB
-Add("▶️ /space/content/add/blob 👩‍💻 🤖")
+Add("⏯️ /space/content/add/blob 👩‍💻 🤖")
+AddOk("🧾 { ok: { claim } }")
 
-Added("🧾 { ok: { claim 🚦 } }")
 
-Allocate("⏭️ /service/blob/allocate 🤖")
+subgraph accept
+Accept("⏯️ /service/blob/accept 🤖")
+AcceptOk("🧾 { ok: { claim } }")
+end
 
-Put("⏭️ /http/put 🤖🔑")
+subgraph put
+Put("⏯️ /http/put 🔑")
+PutOk("🧾 { ok: {} }")
+end
 
-Accept("⏭️ /service/blob/accept 🤖")
 
-Accepted("🧾 { ok: { claim 🎫 } }")
-
+subgraph allocate
+Allocate("⏯️ /service/blob/allocate 🤖")
+AllocateOk("🧾 { ok: { url headers } }")
+end
 
 Claim("🎫 /assert/location 🤖👩‍💻")
 
-Add --o Added
-Added -.-> Allocate
-Added -.-> Accept
-Added -.-> Put
-Allocate --> Accept
 
-Put --> Accept
-Accept --o Accepted
-Accepted -- claim --> Claim
-Added -- claim --> Accept
+
+Add -.-> AddOk
+AddOk -- ⏭️ --> allocate
+AddOk -- ⏭️ --> put
+AddOk -- ⏭️ --> accept
+AddOk -- 🚦 out.ok.claim --> AcceptOk
+
+
+
+Allocate -.-> AllocateOk
+
+Put -- 🚦url --> AllocateOk
+Put -- 🚦headers --> AllocateOk
+
+Put -.-> PutOk
+
+
+
+Accept -- 🚦_put --> PutOk
+Accept -.-> AcceptOk
+AcceptOk -- claim --> Claim
 ```
 
 #### Iconography
 
-- Icon on the left describes type of the node
-- Icon on the right describes issuer / audience pair if only one audience is the issuer
-- ▶️ Task
+- ⏯️ Task
 - ⏭️ Next Task (a.k.a Effect)
 - 🧾 Receipt
-- 🎫 Delegation / Commitment
+- 🎫 Delegation
 - 🚦 Await
 - 👩‍💻 Alice
 - 🤖 Service
+- 🔑 Derived Principal
 
 ### Add Blob Invocation Example
 
@@ -105,9 +123,9 @@ Shown Invocation example illustrates Alice requesting to add 2MiB blob to her sp
   "iss": "did:key:zAlice",
   "aud": "did:web:web3.storage",
   "args": {
-    "blob": {
+    "site": {
       // multihash of the blob as byte array
-      "content": { "/": { "bytes": "mEi...sfKg" } },
+      "digest": { "/": { "bytes": "mEi...sfKg" } },
       // size of the blob in bytes
       "size": 2_097_152,
     }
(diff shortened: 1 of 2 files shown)


w3-blob.md
```diff
@@ -105,9 +123,9 @@ Shown Invocation example illustrates Alice requesting to add 2MiB blob to her sp
   "iss": "did:key:zAlice",
   "aud": "did:web:web3.storage",
   "args": {
-    "blob": {
+    "site": {
```
Gozala, line 126:
```suggestion
    "blob": {
```

w3-blob.md
```diff
@@ -380,15 +426,23 @@ type BlobPut = {
 
 ### Put Blob Subject
 
-The subject field SHOULD be [`did:key`] corresponding to the [Ed25519] private key that is first 32 bytes of the blob [multihash].
+The subject field SHOULD be [`did:key`] corresponding to the [Ed25519] private key that is last 32 bytes of the blob [multihash].
```
Gozala, line 429: @vasco-santos I have changed this to last 32 bytes because that is actual sha-256 of the content while first 32 will include multihash prefix thing which will be always the same.

w3-blob.md
```diff
@@ -418,7 +472,7 @@ Receipt MUST not have any effects.
 
 ## Accept Blob
 
-Authorized agent MAY invoke `/service/blob/accept` capability on the [provider] subject. Invocation MUST either succeed when content is delivered on allocated `address` or fail if either allocation failed or expired before content was delivered.
+Authorized agent MAY invoke `/service/blob/accept` capability on the [provider] subject. Invocation MUST either succeed when content is delivered to an allocated `address` or fail if either allocation failed or expired before content was delivered.
 
 Invocation MUST block until content is delivered. Implementation MAY resume when content is sent to the allocated address or await until client signals that content has been delivered using [put blob receipt].
```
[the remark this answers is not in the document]
vasco-santos: also looking at the first receipt, I think we should have:

```typescript
type BlobAccept = {
  cmd: "/service/blob/accept"
  sub: ProviderDID
  args: {
    blob: Blob
    exp: int
    space: SpaceDID
    _put: [".out.ok.site", Link<HttpPut>]
  }
}
```

is this correct?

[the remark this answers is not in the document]
vasco-santos: actually on `BlobAccept` we have an issue at the moment: 
1. I agree that we need to have `space` as arg, but we have no way to find out the space, unless we pass it via `http/put`. What do you think @Gozala ?

w3-blob.md
```diff
@@ -304,7 +350,7 @@ The `args.blob` field MUST be set to the `Blob` the space is allocated for.
 
 #### Allocation Cause
 
-The `args.cause` field MUST be set to the [Link] for an [Add Blob] task, that caused an allocation.
+The `args.cause` field MUST be set to the [Link] for the [Add Blob] task, that caused an allocation.
 
 ### Allocate Blob Receipt
```
vasco-santos, line 356: See below. Should BlobPut be the following in terms of capability definition:

```ts
type BlobPut = {
  cmd: "/http/put"
  sub: DID
  args: {
    url: [".out.ok.address.url", Link<BlobAlloc>]
    headers: [".out.ok.address.headers", Link<BlobAlloc>]
    body: Blob
  }
  meta: {
    keys: {[key: DID]: bytes}
  }
}
```

w3-blob.md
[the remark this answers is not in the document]
vasco-santos: Also, `blobAccept` handler needs to do set in Assert.location agent DID as audience, not space right?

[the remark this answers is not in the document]
Gozala: No it should be a space, because agent acts on space authority. E.g agent may be CI job authorized to upload content, but not authorized to publish it on e.g bitswap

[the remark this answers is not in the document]
vasco-santos: ok, that seems correct. so then question remains to, how can we get it unless we pass it over on `http/put`

w3-blob.md
Gozala: We talked about this in sync, but capturing the summary here

1. General idea is that arguments are defined in terms of types.
2. Argument can be literals or they can be `ucan/await`  selectors, in which case they effectively act as `await` in js and assume that selector will lead to the value of the expected type.
    - If awaited task fails /  does not have a selected path errors propagate

Unfortunately currently our schema definitions and specifically `.invoke` method on them is unable to recognize `ucan/await` pattern, so short term we could define schema more explicitly using something like

```js
const Await = Schema.struct({
  'ucan/await': Schema.tuple([Schema.text(),  Schema.link()])
})

const Put = Schema.struct({
  can: "http/put",
  with: Schema.did(),
  nb: Schema.struct({
    url: Schema.text().or(Await),
    headers: Schema.dictionary({ value: Schema.string() }).or(Await),
    body: blob
  })
})
```

w3-blob.md
```diff
@@ -418,7 +472,7 @@ Receipt MUST not have any effects.
 
 ## Accept Blob
 
-Authorized agent MAY invoke `/service/blob/accept` capability on the [provider] subject. Invocation MUST either succeed when content is delivered on allocated `address` or fail if either allocation failed or expired before content was delivered.
+Authorized agent MAY invoke `/service/blob/accept` capability on the [provider] subject. Invocation MUST either succeed when content is delivered to an allocated `address` or fail if either allocation failed or expired before content was delivered.
```
Gozala, line 475:
```suggestion
Authorized agent MAY invoke `/service/blob/accept` capability on the [provider] subject. Invocation MUST either succeed when content is delivered at allocated site or fail if either allocation failed or expired before content was delivered.
```

w3-blob.md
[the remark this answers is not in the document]
Gozala: We talked about this in sync and have concluded that we should not capture `space` in the `http/put` because technically there could be multiple `blob/accept`'s across several spaces awaiting for the blob content and it does not matter which one will deliver it. In fact `space` where content is delivered is web3.storage space.

However completing `blob/accept` on other all spaces awaiting is out of scope for the current implementation. That said we can still obtain the space info from the `/http/put` because it's `url` and `headers` args will have `ucan/await` linked to the `/service/blob/allocate` that has `space` argument.

w3-blob.md
```diff
@@ -380,15 +426,23 @@ type BlobPut = {
...
+### Put Blob URL
+
+Destination `url` MUST be specified in the arguments and it MUST be an endpoint that can accept HTTP PUT request.
+
+### Put Blob Headers
+
+The `headers` map MUST be specified in the arguments. It MUST have string keys corresponding to header names and string values corresponding to header values.
 
-### Put Blob Address
+### Put Blob Body
```
Gozala, line 443: Per @alanshaw https://github.com/web3-storage/w3up/pull/1342/files#r1557547248

> Personally I'd just repeat the fields here so this doesn't depend on blob capability.


I think it is a valid point, and I have considered defining schema as following instead

```ts
type BlobPut = {
  cmd: "/http/put"
  sub: DID
  args: {
    url: URL
    headers: Headers
    body: Body
  }
  meta: {
    keys: {[key: DID]: bytes}
  }
}

type Body = Variant<{
  blob: Blob
}>
```

So we could in the future define alternative body types. However I'm not sure if there could be a body that we do not address by hash which is why decided to not introduce another wrapper layer, that said I could be convinced otherwise.

Also relatedly I'm hoping to change `Blob` definition into a multihash that captures payload size per https://github.com/multiformats/multihash/issues/163, which is why `Blob` is treated as a single a unit as opposed to pair of fields.

Gozala: One other thing we have discussed is changing body to just multihash and moving `size` into `headers['content-length']` which we can do if that is preferred. From what I know I don't think presigned URLs currently care about `size` anyway.

vasco-santos: I would really prefer to have the size instead of expecting the size to be in the headers.

w3-blob.md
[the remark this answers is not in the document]
vasco-santos: thanks for summarizing here. 

@Gozala I think the first commend in this thread is still relevant though:

> can't suggest there because is not part of the diff but `blob/accept` receipt type should have site right?

```typescript
type BlobAcceptOk = {
  site: Link<LocationClaim>
}
```

w3-blob.md
```diff
@@ -122,74 +140,102 @@ Shows an example receipt for the above `/space/content/add/blob` capability invo
...
-          // space where memory is allocated
-          "space": "did:key:zAlice",
-          "blob": {
-            // multihash of the blob as byte array
-            "content": { "/": { "bytes": "mEi...sfKg" } },
-            // size of the blob in bytes
-            "size": 2_097_152,
+    "assert": [
+      // refers to the invocation from the example
+      { "/": "bafy..add" },
```
alanshaw, line 151: Can you please add `// "/": "bafy..add"` comment to the previous example?

alanshaw (approved):
LGTM, aside from tweaks and corrections captured in comments.

w3-blob.md
```diff
@@ -51,48 +51,66 @@ Authorized agent MAY invoke `/space/content/add/blob` capability on the [space]
...
-Add("▶️ /space/content/add/blob 👩‍💻 🤖")
+Add("⏯️ /space/content/add/blob 👩‍💻 🤖")
+AddOk("🧾 { ok: { site } }")
 
-Added("🧾 { ok: { claim 🚦 } }")
 
-Allocate("⏭️ /service/blob/allocate 🤖")
+subgraph accept
+Accept("⏯️ /service/blob/accept 🤖")
+AcceptOk("🧾 { ok: { site } }")
```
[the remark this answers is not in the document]
vasco-santos: this is in theory site, but it is not updated below. see https://github.com/web3-storage/specs/pull/118#discussion_r1559008405

w3-blob.md
```diff
@@ -51,48 +51,66 @@ Authorized agent MAY invoke `/space/content/add/blob` capability on the [space]
...
-Put --> Accept
-Accept --o Accepted
-Accepted -- claim --> Claim
-Added -- claim --> Accept
+
+Add -.-> AddOk
+AddOk -- ⏭️ --> allocate
+AddOk -- ⏭️ --> put
+AddOk -- ⏭️ --> accept
+AddOk -- 🚦 out.ok.claim --> AcceptOk
```
Gozala, line 86:
```suggestion
AddOk -- 🚦 out.ok.site --> AcceptOk
```

w3-blob.md
```diff
@@ -51,48 +51,66 @@ Authorized agent MAY invoke `/space/content/add/blob` capability on the [space]
...
-Accept("⏭️ /service/blob/accept 🤖")
-
-Accepted("🧾 { ok: { claim 🎫 } }")
 
+subgraph allocate
+Allocate("⏯️ /service/blob/allocate 🤖")
+AllocateOk("🧾 { ok: { url headers } }")
+end
 
 Claim("🎫 /assert/location 🤖👩‍💻")
```
Gozala, line 78:
```suggestion
Site("🎫 /assert/location 🤖👩‍💻")
```

w3-blob.md
```diff
@@ -51,48 +51,66 @@ Authorized agent MAY invoke `/space/content/add/blob` capability on the [space]
...
+Put -- 🚦url --> AllocateOk
+Put -- 🚦headers --> AllocateOk
+
+Put -.-> PutOk
+
+
+
+Accept -- 🚦_put --> PutOk
+Accept -.-> AcceptOk
+AcceptOk -- claim --> Claim
```
Gozala, line 101:
```suggestion
AcceptOk -- site --> Site
```

vasco-santos (approved):
lgtm

PR #123, opened 2024-04-18
feat: add blob list and remove into spec

vasco-santos:
This PR adds blob listing and removing to the spec as a first version. It is clear we will want to improve this later on, but we agreed in shipping this doing same behaviour as `store/list` and `store/remove` so that we do not lose functionality by shipping blob protocol by default.

With the above in mind, and as we agreed this PR basically ports the `store/list` and `store/remove` here, with some adaptions to be consistent with structure in this document, as well as usage of `ts` code snippets instead of `ipldsch` to also be consistent

### DIFF
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -567,7 +567,7 @@ type LocationCommitment = {
 
 ## List Blob
 
-Authorized agent MAY invoke `blob/list` capability on the [space] subject (`sub` field) to list Blobs added to it at the time of invocation.
+Authorized agent MAY invoke `/space/content/list/blob` capability on the [space] subject (`sub` field) to list Blobs added to it at the time of invocation.
 
 ### List Blob Invocation Example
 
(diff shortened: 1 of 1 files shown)


w3-blob.md
```diff
@@ -563,6 +565,231 @@ type LocationCommitment = {
 }
 ```
 
+## List Blob
+
+Authorized agent MAY invoke `blob/list` capability on the [space] subject (`sub` field) to list Blobs added to it at the time of invocation.
```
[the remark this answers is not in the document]
vasco-santos: oh damn! Yes, sorry this is a typo, I will make it consistent with the `1.0` naming and let a follow up PR to document the current version for all the spec

w3-blob.md
```diff
@@ -563,6 +565,224 @@ type LocationCommitment = {
...
+
+type ListBlobOk = {
+  cursor?: string
+  before?: string
+  after?: string
+  size: number
+  results: ListBlobItem
+}
+
+type ListBlobItem = {
```
Gozala, line 671: Don't want to complicate things here so it's not a request, more of suggestion with an expectation that we'll probably do it sometime in the future instead.

I think we should stop having lists like and instead just return list of receipts instead, we can let client deal with getting / caching receipts locally and deriving timestamps from when receipt was issued.

vasco-santos: Yes, actually @alanshaw and I also talked about shifting into that direction later on, and use more the pattern you described recently on a RFC ("our databases should be considered more an index over our receipts"). But for MVP let's make it like store and iterate on this when possible

PR #126, opened 2024-05-30
feat: add blob/get

joaosa:
Adds the `blob/get` capability spec which is analogous to https://github.com/w3s-project/specs/pull/82

### DIFF
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -823,6 +823,8 @@ Shows an example receipt for the above `/space/content/get/blob` capability invo
     "ran": { "/": "bafy..get" },
     "out": {
       "ok": {
+        // task that caused this invocation
+        "cause": { "/": "bafy..add" }
         "blob": {
           "size": 100,
           "content": { "/": { "bytes": "mEi...sfKg" } },
@@ -867,6 +869,7 @@ type GetBlobReceipt = {
 }
 
 type GetBlobOk = {
+  cause:  Link<Task>
   blob: blob
 }
 
@@ -876,6 +879,10 @@ type GetBlobError = {
   message: string
 }
 ```
+
+#### Get Cause
+
+The `args.cause` field MUST be set to the [Link] for the task, that caused a get.
 
 ##### Get Blob Effects
 
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -886,7 +886,7 @@ The `args.cause` field MUST be set to the [Link] for the task, that caused a get
 
 ##### Get Blob Effects
 
-Receipt MUST not have any effects.
+Receipt MUST NOT have any effects.
 
 # Coordination
 
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -832,7 +832,6 @@ Shows an example receipt for the above `/space/content/get/blob` capability invo
         }
       }
     },
-    // set of tasks to be scheduled.
     "next": []
   }
 }
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -870,7 +870,7 @@ type GetBlobReceipt = {
 
 type GetBlobOk = {
   cause:  Link<Task>
-  blob: blob
+  blob: Blob
 }
 
 type ISO8601Date = string
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -787,7 +787,7 @@ Receipt MUST not have any effects.
 
 Authorized agent MAY invoke `/space/content/get/blob` capability on the [space] subject (`sub` field) to get Blobs added to it at the time of invocation.
 
-This may be used to check for inclusion, or to get the `size` property for a shard.
+This may be used to check for inclusion, or to get the `size` of the blob in bytes.
 
 ### Get Blob Invocation Example
 
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -793,8 +793,10 @@ This may be used to check for inclusion, or to get the `size` of the blob in byt
 
 Shown Invocation example illustrates Alice getting a blob stored on their space.
 
+> ℹ️ We use `// "/": "bafy..` comments to denote CID of the parent object.
+
 ```js
-{
+{ // "/": "bafy..get"
   "cmd": "/space/content/get/blob",
   "sub": "did:key:zAlice",
   "iss": "did:key:zAlice",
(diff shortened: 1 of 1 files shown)


w3-blob.md
```diff
@@ -783,6 +783,108 @@ The `out.ok.size` MUST be set to the number of bytes that were freed from the sp
...
+  "iss": "did:web:web3.storage",
+  "aud": "did:key:zAlice",
+  "cmd": "/ucan/assert/result"
+  "sub": "did:web:web3.storage",
+  "args": {
+    // refers to the invocation from the example
+    "ran": { "/": "bafy..get" },
+    "out": {
+      "ok": {
+        "insertedAt": "2024-04-16T15:49:22.638Z",
```
joaosa, line 826: I kept this around for consistency. Do we actually care about this field? 🤔

vasco-santos (approved):
LGTM

needs origin out though

w3-blob.md
```diff
@@ -783,6 +783,104 @@ The `out.ok.size` MUST be set to the number of bytes that were freed from the sp
...
+#### Get Blob Receipt Schema
+
+```ts
+type GetBlobReceipt = {
+  out: Result<GetBlobOk, GetBlobError>
+  next: []
+}
+
+type GetBlobOk = {
+  blob: blob
```
vasco-santos, line 870: I saw in the implementation you add the `cause` here. And I actually think that is nice, so also should be in the spec response

joaosa: makes perfect sense! adding it

w3-blob.md
```diff
@@ -783,6 +783,104 @@ The `out.ok.size` MUST be set to the number of bytes that were freed from the sp
 
 Receipt MUST not have any effects.
 
+## Get Blob
+
+Authorized agent MAY invoke `/space/content/get/blob` capability on the [space] subject (`sub` field) to get Blobs added to it at the time of invocation.
+
+This may be used to check for inclusion, or to get the `size` property for a shard.
```
alanshaw, line 790:
```suggestion
This may be used to check for inclusion, or to get the `size` of the blob in bytes.
```

w3-blob.md
```diff
@@ -783,6 +783,104 @@ The `out.ok.size` MUST be set to the number of bytes that were freed from the sp
...
+
+### Get Blob Invocation Example
+
+Shown Invocation example illustrates Alice getting a blob stored on their space.
+
+```js
+{
```
alanshaw, line 797: `// "/": "bafy..` is the CID of the object that is the invocation, it is referenced in the receipt. We use these comments so that the receipt (further down) can reference the invocation.

```suggestion

> ℹ️ We use `// "/": "bafy..` comments to denote CID of the parent object.

```js
{ // "/": "bafy..get"
```

w3-blob.md
```diff
@@ -783,6 +783,104 @@ The `out.ok.size` MUST be set to the number of bytes that were freed from the sp
...
+```
+
+### Get Blob Receipt Example
+
+Shows an example receipt for the above `/space/content/get/blob` capability invocation.
+
+> ℹ️ We use `// "/": "bafy..` comments to denote CID of the parent object.
+
+```js
+{ // "/": "bafy..get",
```
alanshaw, line 816: No need for this in the receipt as there is nothing that refers to it later.

```suggestion

```js
{
```

w3-blob.md
```diff
@@ -783,6 +783,104 @@ The `out.ok.size` MUST be set to the number of bytes that were freed from the sp
...
+    "ran": { "/": "bafy..get" },
+    "out": {
+      "ok": {
+        "blob": {
+          "size": 100,
+          "content": { "/": { "bytes": "mEi...sfKg" } },
+        }
+      }
+    },
+    // set of tasks to be scheduled.
```
alanshaw, line 834:
```suggestion
```

w3-blob.md
alanshaw, line 873: There's no type defined for `blob` as far as i can see...

```suggestion
  blob: Blob
```

w3-blob.md
```diff
@@ -783,6 +783,104 @@ The `out.ok.size` MUST be set to the number of bytes that were freed from the sp
...
+
+type GetBlobOk = {
+  blob: blob
+}
+
+type ISO8601Date = string
+
+type GetBlobError = {
+  message: string
+}
```
alanshaw, line 880:
```suggestion
}

type Blob = { /* ??? */ }
```

w3-blob.md
```diff
@@ -783,6 +783,104 @@ The `out.ok.size` MUST be set to the number of bytes that were freed from the sp
...
+```
+
+##### Get Blob Effects
+
+Receipt MUST not have any effects.
```
alanshaw, line 889:
```suggestion
Receipt MUST NOT have any effects.
```

alanshaw (changes requested):
Mostly looks good, just some minor fixes needed.

Gozala:
This might be controversial, but I wonder if we could simply return return pointer to the receipt and let client fetch all the details. The rational here is that adding a blob is a long running task and has state that moves from initiated to in progress to complete described by individual tasks.

I think it would make most sense to do that. That is not to say we should block shipping blob/get on this, but perhaps it would be better to not spec out in way that requires state checks on the server.

Please note that I'm not advocating for the above solution, it's really more of a question than prescription.

Gozala:
In the implementation PR I have proposed adding version suffix to the capability https://github.com/w3s-project/w3up/pull/1484#pullrequestreview-2089148998 I think it would make sense to add this changes to the spec with version suffix also maybe with some note that likely we'll deprecate it in the future in favor of suffixes-less version that just returns receipt id.

vasco-santos (approved):

Gozala (approved):

w3-blob.md
```diff
@@ -783,6 +783,114 @@ The `out.ok.size` MUST be set to the number of bytes that were freed from the sp
...
+  blob: Blob
+}
+
+type ISO8601Date = string
+
+type GetBlobError = {
+  message: string
+}
+
+type Blob = { /* ??? */ }
```
alanshaw, line 883: I think you need to fill this out - it's multihash and size right?

alanshaw (changes requested):

w3-blob.md
joaosa: good point. completely missed it

joaosa: @alanshaw addressed

alanshaw (approved):

PR #134, opened 2025-04-01
fix: blob spec in UCAN 0.9 parlance

alanshaw:
This PR updates the blob spec to use UCAN 0.9 parlance - the version implemented and used in production.

[📖 Preview](https://github.com/storacha/specs/blob/fix/blob-spec-ucan-09-parlance/w3-blob.md)

### DIFF
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -1,6 +1,6 @@
 # W3 Blob Protocol
 
-![status:wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+![reliable](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square)
 
 ## Authors
 
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -912,6 +912,6 @@ Receipt MUST NOT have any effects.
 [DID]:https://www.w3.org/TR/did-core/
 [Link]:https://ipld.io/docs/schemas/features/links/
 [range request]:https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests
-[`did:key`]:https://w3c-ccg.github.io/did-method-key/
+[`did:key`]:https://w3c-ccg.github.io/did-key-spec/
 [Ed25519]:https://en.wikipedia.org/wiki/EdDSA#Ed25519
 [UCAN Conclusion]:./w3-ucan.md#conclusion
--- a/w3-blob.md
+++ b/w3-blob.md
@@ -126,7 +126,7 @@ Shown Invocation example illustrates Alice requesting to add 2MiB blob to her sp
   "aud": "did:web:up.storacha.network",
   "att": [{
     "can": "space/blob/add",
-    "with": "did:key:zAlice",
+    "with": "did:key:zAliceSpace",
     "nb": {
       "blob": {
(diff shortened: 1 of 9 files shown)


w3-blob.md
```diff
@@ -118,127 +118,57 @@ AcceptOk -- site --> Site
...
-  "args": {
-    "blob": {
-      // multihash of the blob as byte array
-      "digest": { "/": { "bytes": "mEi...sfKg" } },
-      // size of the blob in bytes
-      "size": 2_097_152,
+  "aud": "did:web:up.storacha.network",
+  "att": [{
+    "can": "space/blob/add",
+    "with": "did:key:zAlice",
```
volmedo, line 129: `did:key:zAlice` is Alice's key, right? I see the did is also used in a lot of places where a space did is expected, such as the `with` here. Should the with read something like `"with": "did:key:zAliceSpace"`?

If that's the case, the same happens in many other `with`s through the document.

w3-blob.md
```diff
@@ -305,12 +238,12 @@ type AddBlobError = {
...
-1. Provided **sub**ject space is not provisioned with a provider.
+1. Provided subject space is not provisioned with a provider.
 1. Provided `blob.size` is outside of supported range.
 1. Provided `blob.digest` is not a valid [multihash].
 1. Provided `blob.digest` [multihash] hashing algorithm is not supported.
 
-Invocation MUST succeed if non of the above is true. Success value MUST be an object with a `site` field set to [ucan/await] of the task that produces [location commitment].
+Invocation MUST succeed if non of the above is true. Success value MUST be an object with a `site` field set to [ucan/await] of the task that produces a [location commitment].
 
 Task linked from the `site` of the success value MUST be present in the receipt effects _(`next` field)_.
```
volmedo, line 248:
```suggestion
Task linked from the `site` of the success value MUST be present in the receipt effects _(`fx` field)_.
```

w3-blob.md
```diff
@@ -523,106 +540,106 @@ Receipt MUST not have any effects.
...
-    // from this range
-    ["==", ".range[0]", 0],
-    ["==", ".range[1]", 2_097_152],
-  ],
+  "att": [{
+    "can": "assert/location",
+    "with": "did:key:zStorageProvider",
+    "nb": {
+      // space the content was added to
+      "space": { "/": { "bytes": "did:key:zAlice" } },
```
volmedo, line 558: q: why do we need to use `Bytes<SpaceDID>` here instead of just `SpaceDID` as usual?

volmedo (approved):
Thanks for doing this!

I left a comment about Alice DID vs. Alice's space DID, but it's non blocking.

w3-blob.md
alanshaw: In UCAN 0.9, `with` is a URI, so the DID we use here MUST be a string.

UCAN actually encodes `iss` and `aud` DIDs as bytes, so this is consistent with other DIDs encoded in the UCAN. When we implement UCAN 1.0 it'll be consistent with subject `sub` (what `with` becomes).

w3-blob.md
alanshaw: ACK, yes I'll clarify.

PR #141, opened 2025-08-29
feat: API spec for Revocations Check endpoint

### LINKED ISSUE #193: Specify revocations check mechanism
* new spec in our github repo
* it should pass delegations in the caveat (either a CID or the delegation itself)
* it should return any revocations of any of the passed CIDs or delegations (so that the client can independently verify the validity of the revocations)
* this is almost certainly not a capability! probably just an HTTP endpoint

fforbeck:
API specification for `GET /{cid}` endpoint: [📚 Preview](https://github.com/storacha/specs/blob/75d40d74216fa2f84d799607b547c3ce7fec7e78/w3-revocations-check.md)

Closes: https://github.com/storacha/upload-service/issues/193

### DIFF
--- a/w3-revocations-check.md
+++ b/w3-revocations-check.md
@@ -56,7 +56,7 @@ Returns a CAR (Content Addressable aRchive) file containing:
 
 ##### CAR File Structure
 
-The CAR file contains a root block with the following structure (shown in DAG-JSON for readability, but it is **RECOMMENDED** that the root block be encoded as [DAG-CBOR](https://ipld.io/docs/codecs/known/dag-cbor/) for efficiency):
+The CAR file contains a root block with the following structure (shown in DAG-JSON for readability, but it is RECOMMENDED that the root block be encoded as [DAG-CBOR](https://ipld.io/docs/codecs/known/dag-cbor/) for efficiency):
 
 ```json
 {
--- a/w3-revocations-check.md
+++ b/w3-revocations-check.md
@@ -1,0 +1,185 @@
+# W3 Revocations Check Protocol
+
+## Authors
+
+- [Felipe Forbeck](https://github.com/fforbeck)
+
+## Editors
+
+- [Alan Shaw](https://github.com/alanshaw)
+
+## Abstract
+
+The W3 Revocations Check protocol provides a public API endpoint for checking if specific delegation CIDs have been revoked and retrieving cryptographic proof data for trustless verification. This protocol enables clients to verify the validity of UCAN delegations by checking their revocation status.
+
+## Language
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+# Introduction
+
+The revocations check protocol allows clients to verify whether UCAN delegations have been explicitly revoked. The protocol provides a simple HTTP GET endpoint that returns either cryptographic proof of revocation or indicates that no revocation exists for the queried delegation CID.
+
+# Capabilities
+
+## Revocation Check
+
+### Endpoint Details
+
+- **Method:** `GET`
+- **Path:** `/revocations/{cid}`
+- **Authentication:** None required (public endpoint)
+- **CDN Cacheable:** Yes (simple GET requests with 200/404 responses)
+
+### Request Format
+
+#### URL Parameters
+
+**`cid`** (required): Delegation CID to check for revocations. Must be a valid IPFS CID string.
+
+#### Constraints
+
+- **CID Format:** Valid IPFS CID string (e.g., `bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi`)
+- **Encoding:** CID will be normalized to string format internally
+
+### Response Format
+
+#### Success Response (200): Delegation Revoked
+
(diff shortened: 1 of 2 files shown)


w3-revocations-check.md
```diff
@@ -0,0 +1,185 @@
...
+```javascript
+/**
+ * Recursively verify the delegation proof chain for revocations
+ */
+async function verifyProofChain(delegation) {
+  // Check if delegation is explicitly revoked
+  const response = await fetch(`https://up.storacha.network/revocations/${delegation.cid}`);
+  
+  if (response.status === 200) {
+    return { isValid: false, reason: 'Delegation explicitly revoked' };
```
alanshaw, line 172: Note: this is not doing verification, it's _trusting_ the 200 from the service.

Would be worth building a client library that does the verification.

I would perhaps check from the start of the delegation chain first.

w3-revocations-check.md
```diff
@@ -0,0 +1,185 @@
...
+### Request Format
+
+#### URL Parameters
+
+**`cid`** (required): Delegation CID to check for revocations. Must be a valid IPFS CID string.
+
+#### Constraints
+
+- **CID Format:** Valid IPFS CID string (e.g., `bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi`)
+- **Encoding:** CID will be normalized to string format internally
```
alanshaw, line 43:
```suggestion
- **Encoding:** the CID SHOULD be encoded using multibase base32
```

w3-revocations-check.md
```diff
@@ -0,0 +1,185 @@
...
+curl -X GET https://up.storacha.network/revocations/{cid} \
+  -H "Accept: application/vnd.ipld.car" \
+  -o revocation-proof.car
+```
+
+## Client-Side Proof Chain Verification
+
+```javascript
+/**
+ * Recursively verify the delegation proof chain for revocations
```
alanshaw, line 165: I would not...recurse. I'd collect the CIDs we want to check (start of chain first) and make parallel requests (with concurrency limit) allowing cancellation if one is found to be revoked.

I don't think it's worth putting this example in the spec - I'd probably remove.

w3-revocations-check.md
```diff
@@ -0,0 +1,185 @@
...
+ */
+async function verifyProofChain(delegation) {
```
alanshaw, line 167: This is more `isRevoked(delegation)` or `checkRevocation(delegation)`.

I'd maybe return `Result<Unit, Revoked | Failure>`, where `Revoked` can carry information about the delegation that was revoked and the cause.

```ts
interface Revoked extends Failure {
  name: 'Revoked'
  /** The delegation/proof that was revoked */
  delegation: Delegation
  /** The invocation that revoked the delegation */
  cause: Invocation
}
```

w3-revocations-check.md
```diff
@@ -0,0 +1,185 @@
...
+
+The CAR file contains a root block with the following structure:
+
+```json
+{
+  "revocations@0.0.1": {
+    "revocations": [
+      {
+        "delegation": { "/": "[delegationCID]" },
+        "scope": "[scopeDID]",
```
alanshaw, line 67: I don't think we need to include this, as it is derivable from the cause invocation.

Note: scope is the DID of the authority that issued the revocation - either the issuer or audience of the delegation that is under inspection or one of it's proofs.

alanshaw (changes requested):
This looks great, but there's a few minor things to address before mergable.

w3-revocations-check.md
```diff
@@ -0,0 +1,170 @@
...
+
+Returns a CAR (Content Addressable aRchive) file containing:
+
+- Revocation metadata in `revocations@0.0.1` schema
+- Embedded UCAN proof blocks for trustless verification
+- ETag header for caching optimization
+
+##### CAR File Structure
+
+The CAR file contains a root block with the following structure (shown in DAG-JSON for readability, but it is **RECOMMENDED** that the root block be encoded as [DAG-CBOR](https://ipld.io/docs/codecs/known/dag-cbor/) for efficiency):
```
alanshaw, line 59: Capitals is enough 🤣 

```suggestion
The CAR file contains a root block with the following structure (shown in DAG-JSON for readability, but it is RECOMMENDED that the root block be encoded as [DAG-CBOR](https://ipld.io/docs/codecs/known/dag-cbor/) for efficiency):
```

w3-revocations-check.md
```diff
@@ -0,0 +1,170 @@
...
+
+```json
+{
+  "revocations@0.0.1": {
+    "revocations": [
+      {
+        "delegation": { "/": "[delegationCID]" },
+        "scope": "[scopeDID]",
```
alanshaw, line 67:
```suggestion
```

w3-revocations-check.md
```diff
@@ -0,0 +1,170 @@
...
+      }
+    ]
+  }
+}
+```
+
+**Field Descriptions:**
+
+- **`delegation`**: CID of the delegation that has been revoked
+- **`scope`**: DID of the authority that issued the revocation (either the issuer or audience of the delegation under inspection, or one of its proofs). *Note: This field is derivable from the cause invocation but included for convenience*
```
alanshaw, line 78: What is the convenience? I just...can't think of a reason.

```suggestion
```

alanshaw (approved):
After requested changes this LGTM.

travis:
I'm a little late here, but curious if you considered adding a "multi-CID" version of this endpoint - it seems to me like a common usecase will be checking multiple revocations at once (for example, to check if any delegations in a proof chain have been revoked) and that the way this is currently set up that will be quite inefficient... was this an intentional omission?

PR #150, opened 2026-01-22
feat: add plan limit information to plan/get results

volmedo:
Ref. https://github.com/storacha/project-tracking/issues/620

Currently, `plan/get` only returns the name of the plan a given account is subscribed to. This PR adds information about the storage limit for the account, which makes special sense when using the reserved capacity plan in Storacha Forge.

### DIFF
--- a/w3-plan.md
+++ b/w3-plan.md
@@ -26,7 +26,9 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ### `plan/get`
 
-Capability can be invoked to get information about the plan that the account is currently signed up for. The plan name and the storage limit are returned. The plan limit is expressed in bytes. Not to be confused with the storage **included** in the plan. A `0` limit indicates that uploads will not be rejected due to hitting a limit (the plan is effectively "unlimited"), which doesn't mean the customer won't be charged for the extra storage.
+Capability can be invoked to get information about the plan that the account is currently signed up for. The plan name and the storage limit are returned.
+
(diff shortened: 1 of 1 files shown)


w3-plan.md
```diff
@@ -26,7 +26,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ### `plan/get`
 
-Capability can be invoked to get information about the plan that the account is currently signed up for.
+Capability can be invoked to get information about the plan that the account is currently signed up for. The plan name and the storage limit are returned. The plan limit is expressed in bytes. Not to be confused with the storage **included** in the plan. A `0` limit indicates that uploads will not be rejected due to hitting a limit (the plan is effectively "unlimited"), which doesn't mean the customer won't be charged for the extra storage.
```
alanshaw: Optional field that is not present if no limit is set?

alanshaw (approved):

w3-plan.md
volmedo: mmm, I don't know. The code implementing this will end up calling `planLimit` in w3infra and that returns `0` for no limit, so it would require removing it from the response explicitly and is probably not worth it.

alanshaw: I think it's ok for a function that is explicitly called to get a limit to return 0 to signify no limit, however in the invocation response, sending a value when there is none kinda sucks - it's extra bandwdith for nothing and means that you need to look at the spec to interpret what 0 means. Whereas if there is no limit property when there is no limit makes far more sense to me.

It's also _more_ backward compatible with the existing API to have the field optional - all plans in the hot network currently will not send this property.

This doesn't yet exist in the response, I think simply omiting the property is easy enough, especially in JS.

volmedo: ok, ok, but extra bandwidth? It's a request that will be made in response to a human action, i.e. it won't happen at a high rate. And `"limit": 0` doesn't look like huge overhead for being explicit.

And we should discuss what _more_ backward compatible means 😄

Anyways, done!

